PR #5, opened 2022-05-31
fix dependencies for project and tests

vpuri3:


src/SciMLOperators.jl
```diff
@@ -1,28 +1,35 @@
...
+using DiffEqBase
+
 # Misc
-"""
-$(TYPEDEF)
-"""
+
+#"""
+#$(TYPEDEF)
+#"""
```
ChrisRackauckas, line 13: add DocstringExtensions

PR #15, opened 2022-06-01
compose operator and skeleton of matrix-free

vpuri3:


vpuri3:
done with compose. next will do matrix free, and name change (diffeq* -> sciml) in next PR

vpuri3:
tests passing

src/interface.jl
```diff
@@ -1,5 +1,30 @@
...
+    # SciML
+    isconstant::Bool = false
+    islinear::Bool = false
+    issquare::Bool = false
+    iszero::Bool = false
+
+    has_adjoint = false
+    has_ldiv = false
+    has_ldiv! = false
+end
```
ChrisRackauckas, line 26: Does this actually resolve at compile time in context? This form seems like it would be relying too much on constant propagation heuristics and would break easily. Also, it's a non-standard way to do traits. I think we should stick to the standard here.

vpuri3: TBH i have no idea. im just playing around with different styles to see what sticks. good you caught me before i got in too deep.

vpuri3: this isn't relevant to this PR anyway which is about operator composition.

vpuri3: im thinking more about traits here: https://github.com/SciML/SciMLOperators.jl/pull/16

PR #16, opened 2022-06-02
DiffEq --> SciML

vpuri3:
name change. and,

- `operators/basic_operators.jl` -> `basic.jl`
- `operators/common_defaults.jl` -> `common.jl`
- `operators/diffeq_operators.jl` -> `sciml.jl`


vpuri3:
tests passing. time for the massive copy paste job

vpuri3:
I have taken the code from `common_defaults.jl`, and `interface.jl` and divided them among two files:

- `traits.jl` contains default traits. `include`d first
- `interface.jl` contains the default calling interface (`op(u,p,t), op(du,u,p,t)`) `include`d last

vpuri3:
@ChrisRackauckas what's the standard way to do traits?

ChrisRackauckas:
Make them constant functions, i.e. `isconstant(::MyType) = true`. The reason the traits are defined like this is because then they are compile-time resolved, and therefore if they are used to make type decisions then they will allow for dispatch devirtualization (while the other way could lead to dynamic dispatches or uninferred return types)

vpuri3:
can these functions depend on the contents of a type? Something like
```julia
has_ldiv!(L::SciMLMatrixOperator) = has_ldiv!(L.A)
has_ldiv!(D::Diagonal) = true

L = SciMLMatrixOperator(Diagonal(rand(4)))
has_ldiv!(L) # true
```

or would it be better to define
```julia
has_ldiv!(L::SciMLMatrixOperator{T,Atype}) = has_ldiv!(::Type{AType})
```

?

ChrisRackauckas:
Yes. `has_ldiv!(L::SciMLMatrixOperator) = has_ldiv!(L.A)` will then resolve at compile time to just be the result of `has_ldiv!` on the content, and if that's constant that will then also resolve.

vpuri3:
ok ill complete traits as much as possible and push this out

src/interface.jl
```diff
@@ -10,63 +13,56 @@
...
-5. Optional: diagonal, symmetric, etc traits from LinearMaps.jl.
-6. Optional: exp(A). Required for simple exponential integration.
-7. Optional: expmv(A,u,p,t) = exp(t*A)*u and expmv!(v,A::DiffEqOperator,u,p,t)
-   Required for sparse-saving exponential integration.
-8. Optional: factorizations. A_ldiv_B, factorize et. al. This is only required
-   for algorithms which use the factorization of the operator (Crank-Nicholson),
-   and only for when the default linear solve is used.
-=#
+    (L::T)(u, p, t) = (update_coefficients!(L, u, p, t); L * u)
+    (L::T)(du, u, p, t) = (update_coefficients!(L, u, p, t); mul!(du, L, u))
```
ChrisRackauckas, line 38: I think on v1.6 and above you can just dispatch this on `AbstractSciMLOperator`

src/interface.jl
```diff
@@ -10,63 +13,56 @@
...
-isconstant(::AbstractMatrix) = true
-islinear(::AbstractMatrix) = true
-has_adjoint(::AbstractMatrix) = true
-has_mul(::AbstractMatrix) = true
-has_mul!(::AbstractMatrix) = true
-has_ldiv(::AbstractMatrix) = true
-has_ldiv!(::AbstractMatrix) = false
-has_ldiv!(::Union{Diagonal, Factorization}) = true
+function SciMLOperator(L;
+                       update_func=DEFAULT_UPDATE_FUNC,
```
ChrisRackauckas, line 45: This operator treats it as a closed set, and I don't think that's generally safe.

src/utils.jl
```diff
@@ -0,0 +1,12 @@
...
+"""
+We should define `sparse` for these types in `SciMLBase` instead,
+but that package doesn't know anything about sparse arrays yet, so
+we'll introduce a temporary work-around here.
+"""
+function _sparse end
+
+_sparse(L) = sparse(L)
+_sparse(L::SciMLMatrixOperator) = _sparse(L.A)
+_sparse(L::SciMLScaledOperator) = L.λ * _sparse(L.L)
```
ChrisRackauckas, line 11: What exactly is this doing? Why is it not just `SparseArrays.sparse`?

test/basic.jl
```diff
@@ -1,28 +1,29 @@
...
-                      AddedDiffEqOperator,
-                      ComposedDiffEqOperator,
-                      DiffEqFunctionOperator
+using SciMLOperators: SciMLIdentity,
+                      SciMLNullOperator,
+                      SciMLScaledOperator,
+                      SciMLAddedOperator,
+                      SciMLComposedOperator,
+
+                      getops
```
ChrisRackauckas, line 10: what is `getops` for?

src/interface.jl
vpuri3: So we don't need the `update_coefficients` interface? what should be the standard calling interface for SciMLOperators? Just `*, mul!, ldiv!`?

src/utils.jl
vpuri3: preserved it from older code. it's easier to just extend `SparseArrays.sparse`

vpuri3: will rm `_sparse` and do that

test/basic.jl
vpuri3: `getops` is used in the `update_coefficients!` interface like

```julia
function update_coefficients!(L::AbstractSciMLOperator, u, p, t)
    for op in getops(L)
        update_coefficients!(op, u, p, t)
    end
    L
end
```

vpuri3: was thinking of adding tests for `getops` but didn't get around to do it.

src/interface.jl
vpuri3: rm'd

src/interface.jl
ChrisRackauckas: I mean, I think you can just do `(L::AbstractSciMLOperator)(u, p, t) = (update_coefficients!(L, u, p, t); L * u)` on v1.6 and above, and then not have to worry about defining more dispatches of calls.

test/basic.jl
ChrisRackauckas: oh I see, interesting.

src/interface.jl
vpuri3: oh i see what you mean

vpuri3: done

src/SciMLOperators.jl
```diff
@@ -29,70 +22,35 @@ abstract type AbstractSciMLOperator{T} end
...
-export DiffEqScalar,
-       DiffEqArrayOperator,
-       FactorizedDiffEqArrayOperator,
-       AffineDiffEqOperator,
-       MatrixFreeOperator
+export SciMLScalar,
+       SciMLMatrixOperator,
+       SciMLFactorizedOperator,
+       AffineSciMLOperator,
+       SciMLFunctionOperator
```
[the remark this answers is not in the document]
vpuri3: should be ok so long as we don't export these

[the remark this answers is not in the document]
ChrisRackauckas: It would only export if someone `using SciMLOperators`, which would be a pretty deliberate thing that only exports the operator types. I don't think any of the solver libraries will export the basic operators, so this should be fine I think.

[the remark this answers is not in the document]
ChrisRackauckas: Also, I don't think these names would clash with anything. `Scalar` and `Identity` would be bad ideas to export though 😅

src/interface.jl
```diff
@@ -10,63 +13,30 @@
...
+    for op in getops(L)
+        update_coefficients!(op, u, p, t)
+    end
+    L
+end
+
+for T in (
+          SciMLIdentity,
+          SciMLNullOperator,
+          SciMLScalar,
```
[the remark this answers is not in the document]
vpuri3: im also gonna rename `SciMLIdentity` to `SciMLIdentityOperator` just to be clear that this is an operator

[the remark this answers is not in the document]
vpuri3: done

src/SciMLOperators.jl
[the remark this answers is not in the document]
vpuri3: done

vpuri3:
good to go for now or do you want more work on traits in this PR?

vpuri3:
yeah and NullOperator too

vpuri3:
you know what same applies to `ScaledOperator`, `AddedOperator`, and `ComposedOperator`

ChrisRackauckas:
And rebase to the new master so CI runs

PR #19, opened 2022-06-03
FunctionOperator

vpuri3:
need to do tests


vpuri3:
@ChrisRackauckas take a look?

src/sciml.jl
```diff
@@ -255,66 +255,168 @@ end
...
-             op, op_adjoint, op_inverse, size, p,
+             op, op_adjoint, op_inverse, op_adjoint_inverse, traits, p, t,
             )
     end
+
 end
 
-Base.size(L::FunctionOperator) = L.size
-Base.adjoint(L::FunctionOperator) = FunctionOperator(L.op_adjoint; op_inverse=L.op)
+function FunctionOperator(op, traits=nothing;
```
ChrisRackauckas, line 294: It seems like this is the same as the other kwargs below?

vpuri3: it's a named tuple that holds those kwargs.

```julia
# src/sciml.jl
    traits = !(traits isa Nothing) ? traits : (;
                                               opnorm = opnorm,
                                               isreal = isreal,
                                               issymmetric = issymmetric,
                                               ishermitian = ishermitian,
                                               isposdef = isposdef,


                                               isinplace = isinplace,
                                               T = T,
                                               size = size,
                                              )```

Comment (vpuri3) [reply src/sciml.jl:294]: makes it easy to pass things around eg, the definition of adjoint:

```julia
Base.size(L::FunctionOperator) = L.traits.size
function Base.adjoint(L::FunctionOperator{iip,T}) where{iip,T}

    if ishermitian(L) | (isreal(L) & issymmetric(L))
        return L
    end

    op = L.op_adjoint
    op_adjoint = L.op

    op_inverse = L.op_adjoint_inverse
    op_adjoint_inverse = L.op_inverse

    traits = (L.traits[1:end-1]..., size=reverse(size(L)))

    p = L.p
    t = L.t

    FuncitonOperator(op, op_adjoint, op_inverse, op_adjoint_inverse, traits, p, t)
end
```

ChrisRackauckas: But then why have the other keyword arguments? This construction is redundant, unless it's only intended to be an internal and not used/known by users.

vpuri3: the latter - need it for internals

PR #21, opened 2022-06-03
clean up affine operator

vpuri3:
- [x] define traits
- [X] add tests
- [X] mul!, *


vpuri3:
im going with `MuladdOperator` for now

ChrisRackauckas:
>what is the matrix representation of

You can dig up some of the old conversations in DiffEqOperators.jl about this question (https://github.com/SciML/DifferentialEquations.jl/issues/260#issuecomment-370111767, https://github.com/SciML/DiffEqOperators.jl/issues/53). There is no matrix representation of it, since if there was it would be a linear operator instead of an affine operator. You can only represent an affine operator as a linear operator in a dimension of one larger. Basically, the operation is: `[A b] * [u;1]`, so it requires something modified to the input as well. 

While it this seems like it might doom the idea of using matrix-free affine operators, it turns out that it's okay because of an obscure fact that iterative linear solvers are actually compatible with affine operators. Everyone talks about and proves them for linear solvers, but that's actually over-restrictive.

> You can show some nice properties of affine operators, like that the definition given of * will make Q*x=b in an iterative solver converge to Qa*x = b-Qb (so if we actually define a type that does this in Julia, it will work with iterative solvers without a fuss. Just the same thing idea as writing in residual form).

I forget where I wrote the proof of that, some PDE solvers roadmap. But it's rather  straightforward to prove. If `Q` is the affine operator `Q(x) = Ax + c`, then solving `Qx = b` is equivalent to solving `Ax + c = b` or `Ax = b-c`. You do the same thing of plugging the affine operator into the GMRES/CG/etc. proof, just move the affine part to the rhs residual, and show it converges to solving `Ax = b-c`, and so GMRES/CG/etc. solves `Q(x) = c` for an affine operator properly. That same trick then can be used pretty much anywhere you would've had a linear operator to extend the proof to affine operators, so then `exp(A*t)*v` operations via Krylov methods work for A being affine as well, and all sorts of things. So affine operators have no matrix representation but they are still compatible with pretty much any iterative or Krylov method. 

And many boundary conditions give not linear operators but affine operators, hence the existence of all of this.

test/sciml.jl
```diff
@@ -40,6 +40,31 @@ N = 8
...
+
+    L = AffineOperator(MatrixOperator(D), b)
+    @test L \ u ≈ D \ (u - b)
+    #
+    #   TODO uncomment later
+    #   ldiv! for MatrixOperator defined in
+    #   https://github.com/SciML/SciMLOperators.jl/pull/22
+    #
+#   v=rand(N); @test ldiv!(v, L, u) ≈ D \ (u-b)
+#   v=rand(N); @test ldiv!(L, u) ≈ D \ (u-b)
```
ChrisRackauckas, line 65:
```suggestion
v=rand(N); @test ldiv!(v, L, u) ≈ D \ (u-b)
v=rand(N); @test ldiv!(L, u) ≈ D \ (u-b)
```

test/sciml.jl
```diff
@@ -40,6 +40,26 @@ N = 8
...
+    L = AffineOperator(MatrixOperator(A), b)
+
+    @test L * u ≈ A * u + b
+    v=rand(N); @test mul!(v, L, u) ≈ A * u + b
+    v=rand(N); w=copy(v); @test mul!(v, L, u, α, β) ≈ α*(A*u + b) + β*w
+
+    L = AffineOperator(MatrixOperator(D), b)
+    @test L \ u ≈ D \ (u - b)
+v=rand(N); @test ldiv!(v, L, u) ≈ D \ (u-b)
+v=rand(N); @test ldiv!(L, u) ≈ D \ (u-b)
```
ChrisRackauckas, line 60:
```suggestion
    v=rand(N); @test ldiv!(v, L, u) ≈ D \ (u-b)
    v=rand(N); @test ldiv!(L, u) ≈ D \ (u-b)
```

ChrisRackauckas (approved):

Issue #17, opened 2022-06-02
5 argument mul!, 2 argument ldiv!


vpuri3:

danielwe:
Hi, just chiming in to cheer on this project :tada: and agree that, yes, 5-arg mul! is crucial :+1: Not least for the library internals: if all operators support 5-arg mul!, that speeds up and simplifies the mul! implementation for ScaledDiffEqOperator and AddedDiffEqOperator, and the latter no longer needs to preallocate any caches. Both of the following (and their 5-arg generalization) can be replaced by a single call to 5-arg mul! with no reference to `L.cache`:

https://github.com/SciML/SciMLOperators.jl/blob/d58961f6fed1083bf45e101703e0272a788e9582/src/operators/basic_operators.jl#L286-L287

https://github.com/SciML/SciMLOperators.jl/blob/d58961f6fed1083bf45e101703e0272a788e9582/src/operators/basic_operators.jl#L389-L390

For function-based operators, the user would ideally be able to provide a 5-arg function, and the fallback when a 3-arg function is provided could be to preallocate a cache to enable allocation-free 5-arg mul!

(This is probably all obvious. I'm just looking forward to using this package. Don't be surprised if you see some PRs down the road.)

PR #22, opened 2022-06-03
5-argument mul!, 2-arg ldiv!

vpuri3:
https://github.com/SciML/SciMLOperators.jl/issues/17


vpuri3:
ok this is also done.

PR #38, opened 2022-06-05
Start real docs

ChrisRackauckas:


ChrisRackauckas:
@vpuri3 I'm merging to get this out the door quickly, but can you come back around and do a PR that describes the new AbstractSciMLOperator interface in full? I simply copied over the old DiffEqOperator text and write a little bit, but I didn't capture all of the functions. Additionally, instead of just being bullets, we should have docstrings on all of the trait functions and then just reference those docstrings in the documentation to define the interface, similar to how the integrator interface is done (https://diffeq.sciml.ai/stable/basics/integrator/)

vpuri3:
yes

Issue #12, opened 2022-06-01
need lazy adjoint fallback for `AbstractDiffEqLinearOperator`


vpuri3:

danielwe:
Highly recommend taking a look at LinearMaps.jl's handling of adjoints and transposes: https://julialinearalgebra.github.io/LinearMaps.jl/stable/generated/custom/#Adjoints-and-transposes. (It's a very well-designed library, just lacking the updating facilities needed in SciML.)

The bottom line is that `Adjoint <: AbstractMatrix`, which I don't think is what you want. Instead, you probably need your own `SciMLAdjointOperator <: AbstractSciMLLinearOperator` type.

And don't forget transposes :upside_down_face:

ChrisRackauckas:
>Highly recommend taking a look at LinearMaps.jl's handling of adjoints and transposes: https://julialinearalgebra.github.io/LinearMaps.jl/stable/generated/custom/#Adjoints-and-transposes. (It's a very well-designed library, just lacking the updating facilities needed in SciML.)

Agreed, if it didn't do its interface via subtyping I would say we should just be extending it, but its design makes it hard to extend while also making things have a subtype we can dispatch on. Specifically, we'd have to do a lot type pirating to do things like add calls to their types 😅 . So there's probably a lot that we want to copy from there, including the adjoint and transpose ideas.

vpuri3:
@danielwe us there any other functionality you'd like us to copy from LinearMaps?

danielwe:
Thanks so much for asking! As far as my use cases are concerned, the important points are:

* Lazy scaling, composition, and linear combination
* Straightforward updating that recurses down the operator tree
* 5-arg `mul!`
* Non-square operators (even though the full operator is usually N x N in SciML use cases, you want to be able to represent it as a low-rank decomposition (N x 2) x (2 x N))

Just skimming the code briefly but it looks like it's all there :tada:

You also have a lazy tensor product, which is awesome! I'm not using that right now, but I would have given a lot to have it a couple of years ago, and it may come back around in the future. (Tensor sums, block operators, and block diagonal operators would also be nice down the road I suppose, but they're not urgent for me at least.)

[...]

PR #47, opened 2022-06-08
remove eager operator fusion to preserve update behaviour

vpuri3:
ref this thread: https://github.com/SciML/SciMLOperators.jl/issues/12


vpuri3:
@ChrisRackauckas this is good to go

PR #52, opened 2022-06-09
small fixes

vpuri3:
- clarify TensorProductOperator constructor.
- fixed overloads now,  `kron(...) isa AbstractMatrix`, `\otimes(...) isa TensorProductOperator`
- remove redundant +,-,*(SciMLOperator, UniformScaling) methods
- ~exporting `\otimes` as it does not conflict with anything in Base or LinearAlgebra~

### DIFF
--- a/src/SciMLOperators.jl
+++ b/src/SciMLOperators.jl
@@ -47,8 +47,7 @@ export ScalarOperator,
        DiagonalOperator,
        AffineOperator,
        FunctionOperator,
-       TensorProductOperator,
-       ⊗
+       TensorProductOperator
 
 export update_coefficients!,
        update_coefficients,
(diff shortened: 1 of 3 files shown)


src/SciMLOperators.jl
```diff
@@ -46,7 +47,8 @@ export ScalarOperator,
        DiagonalOperator,
        AffineOperator,
        FunctionOperator,
-       TensorProductOperator
+       TensorProductOperator,
+       ⊗
```
ChrisRackauckas, line 51: we don't do unicode in public apis

ChrisRackauckas: bump

vpuri3: done

vpuri3:
good to go

vpuri3:
@ChrisRackauckas good to go

PR #54, opened 2022-06-09
Let SciMLOperators be applied to `AbstractArray`s

vpuri3:


ChrisRackauckas:
It should just be AbstractArray

vpuri3:
TensorProductOperator is left

Issue #64, opened 2022-06-20
`FunctionOperator` FFTW example


vpuri3:
```julia
    ComplexT = if T isa Type{Float16}
        ComplexF16
    elseif T isa Type{Float32}
        ComplexF32     
    else
        ComplexF64
    end

    dx = L / n
    x  = range(start=-L/2, stop=L/2-dx, length=n) |> Array
    
    if T <: Real
        k   = rfftfreq(n, 2π*n/L) |> Array
        tr  = plan_rfft(x)
    else
        k   = fftfreq(n, 2π*n/L) |> Array
        tr  = plan_fft(x)
    end

    transform = FunctionOperator(
                                 (du,u,p,t) -> mul!(du, tr, u);
                                 isinplace=true,
                                 T=ComplexT,
                                 size=(length(k),n),
                                 op_inverse = (du,u,p,t) -> ldiv!(du, tr, u)
                                )

```

ChrisRackauckas:
oh nice. Maybe make that a doc tutorial example?

vpuri3:
playing with it rn. will do a PR soon

Issue #61, opened 2022-06-19
hoist out scaling in tensor product, composition


vpuri3:
```julia
\circ(A::AbstractSciMLOperator, B::ScaleOperator) = B.\lambda * \circ(A, B.L)
```

vpuri3:
any performance advantage given we can't really fuse scalar products...

Issue #66, opened 2022-06-21
potential bug - mistaken operator fusion


vpuri3:
@ChrisRackauckas lazy sum/compose instead?

https://github.com/SciML/SciMLOperators.jl/blob/7f70a6dee148dc2b9ae084fab8a566016399c0a0/src/basic.jl#L226

https://github.com/SciML/SciMLOperators.jl/blob/7f70a6dee148dc2b9ae084fab8a566016399c0a0/src/basic.jl#L232

ChrisRackauckas:
yes lazy unless constant. Are the ScalarOperators not always constant? What's the bug?

vpuri3:
bug = fusion where it should be lazy

ChrisRackauckas:
ahh yes

Issue #63, opened 2022-06-20
Avoid `permutedims!` in `TensorProductOperator` whenever possible


vpuri3:
The two `permutedims!` calls together take about as much time as a single `mul!()`. So we should add more cases to the `L.outer` matvec computation to avoid the `permutedims!` wherever possible.

_Originally posted by @vpuri3 in https://github.com/SciML/SciMLOperators.jl/issues/58#issuecomment-1160455298_

PR #67, opened 2022-06-21
Fix FunctionOperator caching

vpuri3:
- [x] fix caching broken when FunctionOperator isinplace
- [x] close https://github.com/SciML/SciMLOperators.jl/issues/64. ref https://github.com/SciML/SciMLOperators.jl/issues/48
- [x] clean up tensor product operator. add more optimizations - ref https://github.com/SciML/SciMLOperators.jl/issues/63
- [x] close https://github.com/SciML/SciMLOperators.jl/issues/61. ref https://github.com/SciML/SciMLOperators.jl/issues/58
- [x] add benchmarks

 ~https://github.com/SciML/SciMLOperators.jl/issues/66#issuecomment-1161658127~ EDIT - this is a big enough PR as it is. the bug will just have to wait


vpuri3:
running the same test as https://github.com/SciML/SciMLOperators.jl/issues/58#issuecomment-1161928426

```julia
using SciMLOperators, LinearAlgebra, BenchmarkTools

A = TensorProductOperator(rand(12,12), rand(12,12), rand(12,12))

u = rand(12^3, 100)
v = rand(12^3, 100)

A = cache_operator(A, u)

mul!(v, A, u) # dunny
@btime mul!($v, $A, $u); # 4.867 ms (17 allocations: 31.36 KiB)

using SciMLOperators, LinearAlgebra, BenchmarkTools

u = rand(12^2, 100)
v = rand(12^2, 100)

A = rand(12, 12) # outer 
B = rand(12, 12) # inner

T = TensorProductOperator(A, B)
T = cache_operator(T, u)

@btime mul!($v, $T, $u); # 81.514 μs (7 allocations: 304 bytes)

nothing
```
no relevant performance change, though tensor product implementation in this branch is amenable to adding more cases (since `outer_mul, outer_div` can be called recursively).

vpuri3:
need to pass in output prototype to `cache_operator(::FunctionOperator)`

vpuri3:
that doesn't work if the FunctionOperator is nested within a structure that requires caching itself. For eg, if FunctionOperator `F` is composed bw two MatrixOperators, `op = ComposedOperator(A, F, B)` then calling `cache_operator` on `op` would require computing `F*u` which is exactly what we need a cache for when `F` is in place.

much easier to ask for output vector prototype in the constructor of `FunctionOperator`.

vpuri3:
@ChrisRackauckas merge?

PR #80, opened 2022-06-29
Avoid `copy!` scalar indexing error for GPU arrays

vpuri3:
```julia
julia> using CUDA, LinearAlgebra; CUDA.allowscalar(false); u=CUDA.rand(100); v=copy(u); copy!(v,u)
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore are only permitted from the REPL for prototyping purposes.
If you did intend to index this array, annotate the caller with @allowscalar.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] assertscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/rSIl2/src/GPUArraysCore.jl:78
 [3] getindex
   @ ~/.julia/packages/GPUArrays/EVTem/src/host/indexing.jl:9 [inlined]
 [4] copy!(dst::CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}, src::CuArray{Float32, 1, CUDA.Mem.DeviceBuffer})
   @ Base ./abstractarray.jl:886
 [5] top-level scope
   @ REPL[21]:1
 [6] top-level scope
   @ ~/.julia/packages/CUDA/tTK8Y/src/initialization.jl:52

julia> copyto!(v, u)
100-element CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}:
 0.8615155
 0.58178407
 0.98993677
 0.24981835
 0.49794912
 0.91618365
 ⋮
 0.69272184
 0.22738884
 0.9707413
 0.6807149
 0.444413
```


vpuri3:
There are two relevant methods for `copy!`
```julia
[4] copy!(dst::AbstractVector, src::AbstractVector) in Base at abstractarray.jl:881
[5] copy!(dst::AbstractArray, src::AbstractArray) in Base at abstractarray.jl:891
```
The `AbstractVector` one, `[4]`, calls `setindex!`, and `[5]` calls `copyto!`

vpuri3:
```julia
julia> using CUDA, LinearAlgebra; CUDA.allowscalar(false); u=CUDA.rand(100); v=copy(u); copy!(v,u)
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore are only permitted from the REPL for prototyping purposes.
If you did intend to index this array, annotate the caller with @allowscalar.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] assertscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/rSIl2/src/GPUArraysCore.jl:78
 [3] getindex
   @ ~/.julia/packages/GPUArrays/EVTem/src/host/indexing.jl:9 [inlined]
 [4] copy!(dst::CuArray{Float32, 1, CUDA.Mem.DeviceBuffer}, src::CuArray{Float32, 1, CUDA.Mem.DeviceBuffer})
   @ Base ./abstractarray.jl:886
 [5] top-level scope
   @ REPL[14]:1
 [6] top-level scope
   @ ~/.julia/packages/CUDA/tTK8Y/src/initialization.jl:52
```

vpuri3:
```julia
julia> using CUDA, LinearAlgebra; CUDA.allowscalar(false); u=CUDA.rand(10,10); v=copy(u); copy!(v,u)
10×10 CuArray{Float32, 2, CUDA.Mem.DeviceBuffer}:
 0.137139  0.937      0.950817  0.580312    …  0.0964892  0.343289  0.871655   0.434912
 0.359917  0.212419   0.909065  0.976549       0.0701436  0.194903  0.783604   0.667879
 0.886017  0.609401   0.424363  0.162928       0.339217   0.702356  0.281294   0.797848
 0.252522  0.894559   0.761052  0.00161903     0.387095   0.448572  0.774569   0.21113
 0.892149  0.991959   0.319385  0.668126       0.0276097  0.706375  0.639282   0.813365
 0.196853  0.0744156  0.330896  0.94142     …  0.781065   0.285187  0.0224251  0.287546
 0.150622  0.735255   0.874429  0.8663         0.52634    0.514753  0.875352   0.996408
 0.620004  0.033938   0.288622  0.923606       0.818867   0.976576  0.278883   0.304151
 0.287157  0.824294   0.191428  0.354222       0.863757   0.163559  0.332411   0.0735912
 0.696624  0.512505   0.511382  0.193423       0.359115   0.547308  0.18781    0.595577

```

vpuri3:
tests pasing. @ChrisRackauckas this is done

ChrisRackauckas:
Just use `copyto!`?

vpuri3:
copyto! doesn't do size checks so is error prone

ChrisRackauckas:
Why is this not an upstream issue to solve in GPUArrays/CUDA then to add a `copy!` dispatch?

vpuri3:
i created an issue in CUDA but i need a workaround while it is addressed

vpuri3:
would the best workaround be the `_copy!`stuff i put together or just use `copyto!` and add a link to the issue?

ChrisRackauckas:
Wouldn't it take just like 2 minutes to make a PR to GPUArrays.jl that does `copy! = copyto!` on CuArrays? I don't see how this will take so long to get a real fix for. Holding onto upstream stuff like this for small changes is not good for maintenance: just fix the real issue.

vpuri3:
ok wasn't sure about the workflow. will do the PR

ChrisRackauckas:
Generally if the issue is some upstream package not implementing an interface correctly, it's preferred to fix that package. Hardcoding a fix here will still make it not work in other places, so other people will hit it too, and other parts of the code could hit it. So it's almost always recommended to fix upstream. That said, if it looks like this is going to take months to merge or something, then it could make sense to carry a workaround. But this looks like a 5 minute job and others will be happy, so to me it's clear `copy!` for GPUs should just fixed and no one then needs a workaround.

vpuri3:
understood. thanks for explaining. GPUArrays PR is here: https://github.com/JuliaGPU/GPUArrays.jl/pull/414

PR #85, opened 2022-07-07
reshape of reshape is not a strided array

vpuri3:
bug fix

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,6 +1,8 @@
 #
 """ use Base.ReshapedArray """
 _reshape(a, dims::NTuple{D,Int}) where{D} = reshape(a,dims)
+_reshape(a::ReshapedArray, dims::NTuple{D,Int}) where{D} = _reshape(a.parent, dims)
+
 function _reshape(a::AbstractArray, dims::NTuple{D,Int}) where{D}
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -1,6 +1,8 @@
 #
 """ use Base.ReshapedArray """
 _reshape(a, dims::NTuple{D,Int}) where{D} = reshape(a,dims)
+_reshape(a::ReshapedArray, dims::NTuple{D,Int}) where{D} = _reshape(a.parent, dims)
```
ChrisRackauckas, line 4: This is piracy

vpuri3: no because `_reshape` is a method defined in `SciMLOperators`, and is not exported either

per https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy, ""Type piracy" refers to the practice of extending or redefining methods in Base or other packages on types that you have not defined."

`_reshape` is used to get around https://github.com/JuliaLang/julia/issues/36313.  There's reasonable improvement in https://github.com/JuliaLang/julia/pull/43724 and https://discourse.julialang.org/t/convert-array-into-matrix-in-place/55624/3 with this change.

PR #88, opened 2022-07-19
bug fix

vpuri3:
The methods with `Base.-` were incorrect (and not needed) since `-(::AbstractSciMLOperator) isa ScaledOperator`, and that has no field `ops`


vpuri3:
@ChrisRackauckas

vpuri3:
and version bump pls

PR #90, opened 2022-07-22
replace copy with similar for performance

vpuri3:


ChrisRackauckas:
For more safety, we should probably use `zero`.

vpuri3:
@ChrisRackauckas merge?

PR #93, opened 2022-07-27
fix addvector

vpuri3:
tests passing


vpuri3:
@ChrisRackauckas merge?

PR #96, opened 2022-08-08
allow functionoperator to specify both in-place/ out-of-place mode

vpuri3:
with this PR, one can provide both in-place, and out-of-place function signatures to FunctionOperator
```julia
function fwd(v, u, p, t) # method 1
# modify v
end

function fwd(u, p, t) # method 2
...
return v
end

op = FunctionOperator(fwd, isinplace=true, outofplace=true...)

op(v, u, p, t) # uses method 1
v = op(u, p, t) # uses method 2
```


src/func.jl
```diff
@@ -2,7 +2,7 @@
 """
     Matrix free operators (given by a function)
 """
-mutable struct FunctionOperator{isinplace,T,F,Fa,Fi,Fai,Tr,P,Tt,C} <: AbstractSciMLOperator{T}
+mutable struct FunctionOperator{iip,oop,T<:Number,F,Fa,Fi,Fai,Tr,P,Tt,C} <: AbstractSciMLOperator{T}
```
ChrisRackauckas, line 5: this only needs a single parameter

vpuri3:
aahhh, so i wanted to use the same `FunctionOperator` for both in-place, and out-of-place (to play nice with Zygote).

rn on master, when you specify `isinplace=true`, `*, \` use the in-place version
https://github.com/SciML/SciMLOperators.jl/blob/0b4902441115ddf71c2e853a6bf788e05049e2c4/src/func.jl#L266-L276

so i want a trait that says if you have the out-of-place version available, use that instead

vpuri3:
^let me know if that's ok.

In any case, I found a bunch of ways to clean up the API to function operator so ill add it here

vpuri3:
@ChrisRackauckas take a look whenever you have time

Project.toml
```diff
@@ -1,7 +1,8 @@
 name = "SciMLOperators"
 uuid = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"
 authors = ["xtalax <alex.lemony@gmail.com>"]
-version = "0.1.12"
+
```
ChrisRackauckas, line 4:
```suggestion
```

src/func.jl
```diff
@@ -62,43 +66,53 @@ mutable struct FunctionOperator{isinplace,T,F,Fa,Fi,Fai,Tr,P,Tt,C} <: AbstractSc
...
+        static_hasmethod(op, typeof((input, p, t)))
+    else
+        outofplace
     end
 
+    if !isinplace & !outofplace
+        @error "Please provide a funciton with signatures `op(u, p, t)` for applying
+        the operator out-of-place, and/or the signature is `op(du, u, p, t)` for
+        in-place application. Further, we assume that the function call would be
+        nonallocating when called in-place"
```
ChrisRackauckas, line 111: This assumption isn't required?

vpuri3: correct

ChrisRackauckas:
we can try this.

Issue #108, opened 2022-10-04
Dependency on StaticArraysCore instead of StaticArrays?


oschulz:
Could SciMLOperators depend on StaticArraysCore instead of StaticArrays, or does it need full StaticArrays functionality?

ChrisRackauckas:
It could do core.

oschulz:
Oh, that would be neat!

oschulz:
Merci @ChrisRackauckas !

PR #111, opened 2022-10-26
switch to staticarrayscore, rm base.reshapedarray

vpuri3:
close https://github.com/SciML/SciMLOperators.jl/issues/108


vpuri3:
@ChrisRackauckas

PR #122, opened 2022-12-16
Fix op-not-defined error while trying to print error message

gaurav-arya:


vpuri3:
@ChrisRackauckas can you approve please?

Issue #127, opened 2022-12-29
Bugs with applying tensor product operator


gaurav-arya:
Just making a note of some bugs with the tensor product operators: (I know it's an in-progress library :)

- 5-arg `mul!` is broken when the batch size `K` is 1
~- Division via `\` (and likely `ldiv!`) is broken in a number of cases~ (Semi-resolved: these cases were non-square.)

Examples of both can be found in the `@test_broken`'d added in #126.

P.S: It's very tempting to directly define 
```julia
inv(L::TensorProductOperator) = 
      TensorProductOperator(InvertedOperator(L.outer), InvertedOperator(L.inner))
```
and then get `ldiv!` and `\` for free by applying this operator, instead of having to write separate logic for those. I tried to push this through, but it didn't quite work out because for `ldiv!` to be efficient you would want to reuse the cache of `L`, but the cache as currently defined isn't quite symmetric between input and output so this becomes a little tricky. (Maybe #84 helps with this approach?)

vpuri3:
what are the error messages?

gaurav-arya:
For the division error:
```
 Test threw exception
  Expression: AB \ v2 ≈ opAB \ v2 ≈ opAB_F \ v2
  DimensionMismatch: new dimensions (11, 5) must be consistent with array size 21
```

For the `mul!` error:
```
 Test threw exception
  Expression: mul!(v3, opABC, u3, α, β) ≈ α * ABC * u3 + β * w3
  StackOverflowError:
```
(The broken tests for making these errors are in #126)

vpuri3:
and is this on current mastro branch, or on? https://github.com/SciML/SciMLOperators.jl/pull/126?

gaurav-arya:
The same operations error on master (they don't use factorizations, just something I discovered while expanding out the tests)

gaurav-arya:
E.g. here's a MWE that runs on master with all the broken tests:
```
using SciMLOperators, LinearAlgebra
using SciMLOperators: InvertibleOperator, ⊗
using Test
@testset "TensorProductOperator" begin for K in [1, 12]
    m1, n1 = 3 , 5
    m2, n2 = 7 , 11
    m3, n3 = 13, 17

    A = rand(m1, n1)
    B = rand(m2, n2)
    C = rand(m3, n3)
    α = rand()
    β = rand()

    AB  = kron(A, B)
    ABC = kron(A, B, C)

    # Inputs
    u2 = rand(n1*n2, K)
[...]

PR #131, opened 2022-12-29
fix nonsquare Base.:\ for tensor product

vpuri3:
fix https://github.com/SciML/SciMLOperators.jl/issues/127


ChrisRackauckas (approved):

vpuri3:
This is a good stopping point. @ChrisRackauckas this is done

ChrisRackauckas:
I'm not checking it in detail but it looks reasonable.

PR #135, opened 2022-12-29
Initial implementation of InvertedScalarOperator

### LINKED ISSUE #134: Inverted scalar operators do not compose nicely
This one is a little trickier than #132. MWE:
```julia
julia> using SciMLOperators

julia> γ = ScalarOperator(3)
ScalarOperator{Int64, typeof(SciMLOperators.DEFAULT_UPDATE_FUNC)}(3, SciMLOperators.DEFAULT_UPDATE_FUNC)

julia> A = MatrixOperator(rand(3,3))
MatrixOperator{Float64, Matrix{Float64}, typeof(SciMLOperators.DEFAULT_UPDATE_FUNC)}([0.765305882632816 0.678138158666172 0.7379545164098689; 0.48389582113475826 0.4444001494855945 0.7708180654175514; 0.9614440536654981 0.8610110422499186 0.7644354084546402], SciMLOperators.DEFAULT_UPDATE_FUNC)

julia> 1/γ * A
ERROR: BoundsError: attempt to access Tuple{} at index [2]
Stacktrace:
 [1] getindex(t::Tuple, i::Int64)
   @ Base ./tuple.jl:29
 [2] size
   @ ~/.julia/dev/SciMLOperators/src/interface.jl:71 [inlined]
 [3] ComposedOperator
   @ ~/.julia/dev/SciMLOperators/src/basic.jl:429 [inlined]
 [4] #ComposedOperator#79
   @ ~/.julia/dev/SciMLOperators/src/basic.jl:441 [inlined]
 [5] ComposedOperator
   @ ~/.julia/dev/SciMLOperators/src/basic.jl:439 [inlined]
 [6] ∘
   @ ~/.julia/dev/SciMLOperators/src/basic.jl:445 [inlined]
 [7] *(A::SciMLOperators.ScaledOperator{Int64, ScalarOperator{Int64, typeof(SciMLOperators.DEFAULT_UPDATE_FUNC)}, SciMLOperators.InvertedOperator{Int64, ScalarOperator{Int64, typeof(SciMLOperators.DEFAULT_UPDATE_FUNC)}, Nothing}}, B::MatrixOperator{Float64, Matrix{Float64}, typeof(SciMLOperators.DEFAULT_UPDATE_FUNC)})
   @ SciMLOperators ~/.julia/dev/SciMLOperators/src/basic.jl:453
 [8] top-level scope
   @ REPL[57]:1
```
The issue is that `1/γ` is an `InvertedOperator` (not `<: AbstractSciMLScalarOperator`), so the special scalar op handling doesn't apply and things fail since it doesn't have a size. Would an `InvertedScalarOperator` be the simplest way of fixing this problem? With the current design, the only alternative I can see to that would be to deal with `Union{AbstractSciMLScalarOperator, InvertedOperator{T, <:AbstractSciMLScalarOperator} where {T}}` everywhere or equivalently dispatch on a trait.

gaurav-arya:
There's probably some more to fill in, and definitely more to test, but resolves #134 for now. This PR is done on top of #133

### DIFF
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -199,14 +199,16 @@ for T in SCALINGNUMBERTYPES
         ScaledOperator(λ, L.L)
     end
     
-    @eval Base.:*(λ::$T, L::AbstractSciMLOperator) = ScaledOperator(λ, L)
-    @eval Base.:*(L::AbstractSciMLOperator, λ::$T) = ScaledOperator(λ, L)
+    for LT in SCALINGCOMBINETYPES
+        @eval Base.:*(λ::$T, L::$LT) = ScaledOperator(λ, L)
+        @eval Base.:*(L::$LT, λ::$T) = ScaledOperator(λ, L)
 
-    @eval Base.:\(λ::$T, L::AbstractSciMLOperator) = ScaledOperator(inv(λ), L)
-    @eval Base.:\(L::AbstractSciMLOperator, λ::$T) = ScaledOperator(λ, inv(L))
+        @eval Base.:\(λ::$T, L::$LT) = ScaledOperator(inv(λ), L)
+        @eval Base.:\(L::$LT, λ::$T) = ScaledOperator(λ, inv(L))
 
-    @eval Base.:/(L::AbstractSciMLOperator, λ::$T) = ScaledOperator(inv(λ), L)
-    @eval Base.:/(λ::$T, L::AbstractSciMLOperator) = ScaledOperator(λ, inv(L))
+        @eval Base.:/(L::$LT, λ::$T) = ScaledOperator(inv(λ), L)
+        @eval Base.:/(λ::$T, L::$LT) = ScaledOperator(λ, inv(L))
+    end
 end
 
 Base.:-(L::AbstractSciMLOperator) = ScaledOperator(-true, L)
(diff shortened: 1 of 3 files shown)


vpuri3 (approved):

ChrisRackauckas:
What's the reason to not just define the inverse as another scalar operator with inv(x)?

gaurav-arya:
I followed the lazy style of the already existing `AddedScalarOperator`, `ComposedScalarOperator`, etc. -- I'm not sure what benefits there are to it?

gaurav-arya:
Ah, one reason could be that it plays more nicely with update funcs?

gaurav-arya:
The style you're suggesting seems to be followed by the implementation of `conj` for a scalar operator. But that has the disadvantage that if the original op gets updated the conjugated one won't know about it -- there's no backref. That seems like the wrong behavior to me

It's possible that there's a good way of efficiently compressing operators in a non-lazy fashion, without losing their update funcs (in fact, I had a case where I needed this when I wanted to compress a lazily composed operator into a matrix operator but preserve the update func behavior, I should make an issue tomorrow), but it might require some thought.

ChrisRackauckas:
We could define the update_func to put a closure over that then does `1/update_func(u,p,t)`. 

But let's go with this for now. It could be deprecated later.

PR #138, opened 2023-01-11
Prepare for `AdjointFactorization`, multiple dispatch

dkarrasch:
This prepares the ground for https://github.com/JuliaLang/julia/pull/46874. The const definition and the method signatures are harmless for older Julia versions. I also rearranged the `convert` functions and used multiple dispatch. I don't know why, but the changes fix a broken test on Julia v1.9, but not older.

### DIFF
--- a/test/matrix.jl
+++ b/test/matrix.jl
@@ -231,7 +231,12 @@ for square in [false, true] #for K in [1, K]
         v2=copy(u2); @test ldiv!(opAB_F , u2) ≈ AB  \ v2
         v3=copy(u3); @test ldiv!(opABC_F, u3) ≈ ABC \ v3
     else # TODO
-        u2=rand(N2,K); @test_broken ldiv!(u2, opAB_F , v2) ≈ AB  \ v2 # fails
+        u2=rand(N2,K); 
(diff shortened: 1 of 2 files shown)


test/matrix.jl
```diff
@@ -231,7 +231,12 @@ for square in [false, true] #for K in [1, K]
         v2=copy(u2); @test ldiv!(opAB_F , u2) ≈ AB  \ v2
         v3=copy(u3); @test ldiv!(opABC_F, u3) ≈ ABC \ v3
     else # TODO
-        u2=rand(N2,K); @test_broken ldiv!(u2, opAB_F , v2) ≈ AB  \ v2 # fails
+        u2=rand(N2,K); 
+        if VERSION < v"1.9-"
+            @test_broken ldiv!(u2, opAB_F , v2) ≈ AB  \ v2
+        else
+            @test ldiv!(u2, opAB_F , v2) ≈ AB  \ v2 # fails
```
ChrisRackauckas, line 238:
```suggestion
            @test ldiv!(u2, opAB_F , v2) ≈ AB  \ v2
```

PR #143, opened 2023-01-31
Propagate kwargs through update_coefficients!

### LINKED ISSUE #125: Propagate `kwargs` through `update_coefficients!`
In some rare cases, an operator depends on more state than just `u`, `p`, and `t`. For example, the `W` operator in `OrdinaryDiffEq` depends on `dtgamma`. 

If `kwargs` could be propagated through `update_coefficients!`, we'd be able to take advantage of the recursive infrastructure in these cases too. The call might look something like `update_coefficients!(W, u, p, t; dtgamma)`: operators which do not use `dtgamma` would just ignore the extra info, while an operator which would like to use it catches it and uses it. Of course, if an operator requires `dtgamma` (or contains an operator that does), it could not be used as a regular `SciMLOperator` for the usual use cases, but it wouldn't make sense for that to be possible in any case if the operator needs extra state.

For this to work, it seems like all that would be needed to be done is:

- Recursively propagate `kwargs` in `update_coefficients!` for `ComposedOperator`'s, etc.
- User-provided `update_func` signatures which do not catch `kwargs` should be detected and modified to catch `kwargs` (and not use them), so that the usual use case of `update_coefficients!` still works seamlessly. (Something like https://stackoverflow.com/a/42579941?)

In general, something like this just seems like good design: while the positional args `u, p, t` capture 99% of use cases, there are definitely cases (like the `W` operator) which could naturally be expressed as a SciMLOperator if we just allowed an extra state dependence. Happy to take a stab at this if it's deemed a good idea.

gaurav-arya:
Resolves #125. For example usage, see the added test for a scalar operator. ~I left function operator mostly alone because I don't understand some aspects of it, and from some of the other issues it seems it still might be in flux. More comprehensive tests can be added in a follow up PR, but note that this change is totally backwards compatible.~ (Addressed)

For developers of this library, this is what you should know about the new design:

- Operator constructors called by the user should accept `update_func` and `accepted_kwargs` arguments, and preprocess them using the `preprocess_update_func` utility before making the struct, as done here.
- For internal calls to constructors where the `update_func` has already been processed, `accepted_kwargs` should be set to `NoKwargFilter()`, which indicates that no kwarg filtering should be done.

### DIFF
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -61,4 +61,4 @@ matrix-free representations, hence their support in the SciMLOperators interface
 In rare cases, an operator may be used in a context where additional state is expected to be provided
 to `update_coefficients!` beyond `u`, `p`, and `t`. In this case, the operator may accept this additional
 state through arbitrary keyword arguments to `update_coefficients!`. When the caller provides these, they will be recursively propagated downwards through composed operators just like `u`, `p`, and `t`, and provided to the operator.
-For the [premade SciMLOperators](premade_operators.md), one can specify the keyword arguments used by an operator with an `accepted_kwarg_fields` argument that defaults to an empty tuple.
+For the [premade SciMLOperators](premade_operators.md), one can specify the keyword arguments used by an operator with an `accepted_kwargs` argument that defaults to an empty tuple.
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -61,7 +61,7 @@ matrix-free representations, hence their support in the SciMLOperators interface
 In rare cases, an operator may be used in a context where additional state is expected to be provided
 to `update_coefficients!` beyond `u`, `p`, and `t`. In this case, the operator may accept this additional
 state through arbitrary keyword arguments to `update_coefficients!`. When the caller provides these, they will be recursively propagated downwards through composed operators just like `u`, `p`, and `t`, and provided to the operator.
-For the [premade SciMLOperators](premade_operators.md), one can specify the keyword arguments used by an operator with an `accepted_kwargs` argument that defaults to an empty tuple.
+For the [premade SciMLOperators](premade_operators.md), one can specify the keyword arguments used by an operator with an `accepted_kwargs` argument (by default, none are passed).
 
 In the below example, we create an operator that gleefully ignores `u`, `p`, and `t` and uses its own special scaling.
 ```@example
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -55,3 +55,10 @@ the proof to affine operators, so then ``exp(A*t)*v`` operations via Krylov meth
 affine as well, and all sorts of things. Thus affine operators have no matrix representation but they 
 are still compatible with essentially any Krylov method which would otherwise be compatible with
 matrix-free representations, hence their support in the SciMLOperators interface.
+
+## Note about keyword arguments to `update_coefficients!`
+
+In rare cases, an operator may be used in a context where additional state is expected to be provided
+to `update_coefficients!` beyond `u`, `p`, and `t`. In this case, the operator may accept this additional
+state through arbitrary keyword arguments to `update_coefficients!`. When the caller provides these, they will be recursively propagated downwards through composed operators just like `u`, `p`, and `t`, and provided to the operator.
+For the [premade SciMLOperators](premade_operators.md), one can specify the keyword arguments used by an operator with an `accepted_kwarg_fields` argument that defaults to an empty tuple.
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -62,3 +62,19 @@ In rare cases, an operator may be used in a context where additional state is ex
 to `update_coefficients!` beyond `u`, `p`, and `t`. In this case, the operator may accept this additional
 state through arbitrary keyword arguments to `update_coefficients!`. When the caller provides these, they will be recursively propagated downwards through composed operators just like `u`, `p`, and `t`, and provided to the operator.
 For the [premade SciMLOperators](premade_operators.md), one can specify the keyword arguments used by an operator with an `accepted_kwargs` argument that defaults to an empty tuple.
+
+In the below example, we create an operator that gleefully ignores `u`, `p`, and `t` and uses its own special scaling.
+```@example
+using SciMLOperators
+
+γ = ScalarOperator(0.0; update_func=(a, u, p, t; my_special_scaling) -> my_special_scaling,
+                   accepted_kwargs=(:my_special_scaling,))
+
+# Update coefficients, then apply operator
+update_coefficients!(γ, nothing, nothing, nothing; my_special_scaling=7.0)
+@show γ * [2.0]
+
+# Use operator application form
+@show γ([2.0], nothing, nothing; my_special_scaling = 5.0)
+nothing # hide
+```
--- a/src/func.jl
+++ b/src/func.jl
@@ -208,7 +208,7 @@ function update_coefficients!(L::FunctionOperator, u, p, t; kwargs...)
 
     L.p = p
     L.t = t
-    L.kwargs = normalize_kwargs(kwargs)
+    L.kwargs = kwargs
 
     nothing
 end
--- a/src/func.jl
+++ b/src/func.jl
@@ -238,9 +238,6 @@ function Base.adjoint(L::FunctionOperator)
     traits = L.traits
     @set! traits.size = reverse(size(L))
 
-    p = L.p
-    t = L.t
-
     cache = if iscached(L)
         cache = reverse(L.cache)
     else
@@ -252,8 +249,9 @@ function Base.adjoint(L::FunctionOperator)
                      op_inverse,
                      op_adjoint_inverse,
                      traits,
-                     p,
-                     t,
+                     L.p,
+                     L.t,
+                     L.kwargs,
                      cache,
                     )
 end
@@ -280,9 +278,6 @@ function Base.inv(L::FunctionOperator)
         (p::Real) -> 1 / traits.opnorm(p)
     end
 
-    p = L.p
-    t = L.t
-
     cache = if iscached(L)
         cache = reverse(L.cache)
     else
@@ -294,8 +289,9 @@ function Base.inv(L::FunctionOperator)
                      op_inverse,
                      op_adjoint_inverse,
                      traits,
-                     p,
-                     t,
+                     L.p,
+                     L.t,
+                     L.kwargs,
                      cache,
                     )
 end
--- a/src/func.jl
+++ b/src/func.jl
@@ -176,8 +176,7 @@ function FunctionOperator(op,
                          traits,
                          p,
                          t,
-                         # automatically convert NamedTuple's to pairs 
-                         Base.Pairs{Symbol}(kwargs_for_op, keys(kwargs_for_op)),
+                         normalize_kwargs(kwargs_for_op),
                          cache
                         )
 
@@ -209,7 +208,7 @@ function update_coefficients!(L::FunctionOperator, u, p, t; kwargs...)
 
     L.p = p
     L.t = t
-    L.kwargs = kwargs
+    L.kwargs = normalize_kwargs(kwargs)
 
     nothing
 end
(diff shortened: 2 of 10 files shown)


vpuri3:
I like the idea. We can simply pass AD-able parameters in `p`, which can be array-like, and operator relevant context or struct objects in `kwargs`. The latter may contain mesh information, numerical parameters, discretization information. This solving the  primary problem in https://github.com/SciML/DifferentialEquations.jl/issues/881.

vpuri3:
You can go ahead and modify `FunctionOperator`. I'll hold off on any changes for the meantime.

vpuri3:
Would be good if you can modify the `operator algebra` test in `test/total.jl` and have `D1, D2, T1, T2` take in kwargs.

gaurav-arya:
Sure, working on it

gaurav-arya:
Regarding https://github.com/SciML/DifferentialEquations.jl/issues/881: there's still more design work to do in terms of how users can specify extra parameters, but yes hopefully this comes in handy. The nice thing about this PR is that it's very general. All it is doing is making SciMLOperators more flexible: if you want to use a SciMLOperator in a context where something else is expected beyond u/p/t, you can do it. 

It's still up to upstream projects how to use this flexibility: as one possibility, SciML-verse problems could take in an `operator_state` keyword argument for non-differentiable, fixed, parameters. If the user were to make use of this keyword argument, they would be expected to write SciMLOps / DiffEq funcs that accept the `operator_state` kwarg, and SciML would indeed provide it to them.

test/total.jl
```diff
@@ -73,28 +73,51 @@ end
...
     α = rand()
     β = rand()
 
     T1 = ⊗(A, B)
     T2 = ⊗(C, D)
 
-    D1  = DiagonalOperator(rand(N2))
-    D2  = DiagonalOperator(rand(N2))
+    # Introduce update function for D1
+    D1  = DiagonalOperator(p * ones(N2); update_func=(A, u, p, t) -> (A .= p))
```
vpuri3, line 97: can you rename dummy variable `A` to `d` to indicate that it is a vector. Uppercase notation is usually for matrices.

vpuri3: Can you also initialize the operator with zeros like `DiagonalOperator(zeros(N); update_func...)` (currently you are initializing with the right answer). This is so we can concretely see `update_coeffs` in action

test/total.jl
```diff
@@ -73,28 +73,51 @@ end
...
+    # Introduce update funcion for D2 dependent on kwarg "diag" 
+    D2  = DiagonalOperator(p*t * diag; update_func=(A, u, p, t; diag) -> (A .= p*t*diag),
```
vpuri3, line 99: same comments as line 97

test/total.jl
```diff
@@ -73,28 +73,51 @@ end
...
+    dense_DD = Diagonal(vcat(p * ones(N2), p*t*diag))
+    dense_op = hcat(dense_T1', dense_T2') * dense_DD * vcat(dense_T1, dense_T2)
+    # Test correctness of op
+    @test op * u ≈ dense_op * u
+    @test convert(AbstractMatrix, op) ≈ dense_op 
+    # Test consistency with three-arg mul!
     v=rand(N2,K); @test mul!(v, op, u) ≈ op * u
+    # Test consistency with in-place five-arg mul!
     v=rand(N2,K); w=copy(v); @test mul!(v, op, u, α, β) ≈ α*(op * u) + β * w
 end
```
vpuri3, line 122: Can you also add a test for `op(u, p, t; kw..)` and `op(v, u, p, t; kw..)`?

test/scalar.jl
```diff
@@ -82,5 +82,16 @@ end
     @test num(v,u,p,t) ≈ val * u
 
     @test convert(Number, num) ≈ val
+
+    # Test scalar operator which expects keyword argument to update, modeled in the style of a DiffEq W-operator.
+    γ = ScalarOperator(0.0; update_func=(args...; dtgamma) -> dtgamma, accepted_kwarg_fields=(:dtgamma,))
+
+    dtgamma = rand()
+    @test γ(u,p,t; dtgamma) ≈ dtgamma * u
```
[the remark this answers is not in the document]
vpuri3: @gaurav-arya ^

docs/src/interface.md
```diff
@@ -55,3 +55,10 @@ the proof to affine operators, so then ``exp(A*t)*v`` operations via Krylov meth
 affine as well, and all sorts of things. Thus affine operators have no matrix representation but they 
 are still compatible with essentially any Krylov method which would otherwise be compatible with
 matrix-free representations, hence their support in the SciMLOperators interface.
+
+## Note about keyword arguments to `update_coefficients!`
+
+In rare cases, an operator may be used in a context where additional state is expected to be provided
+to `update_coefficients!` beyond `u`, `p`, and `t`. In this case, the operator may accept this additional
+state through arbitrary keyword arguments to `update_coefficients!`. When the caller provides these, they will be recursively propagated downwards through composed operators just like `u`, `p`, and `t`, and provided to the operator.
+For the [premade SciMLOperators](premade_operators.md), one can specify the keyword arguments used by an operator with an `accepted_kwargs` argument that defaults to an empty tuple.
```
vpuri3, line 64: Thanks for the details. Would be good to add an example here once the rest is worked out.

test/scalar.jl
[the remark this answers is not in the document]
vpuri3: It would be nice to be able to provide a default value as well.
```julia
D = DiagonalOperator(zeros(N); update_func=(d, u, p, t; diag = zeros(N)) -> (A .= p * t * diag), accepted_kwargs = (:diag,))

D(u, p, t; diag = diag)
D(u, p, t)
```

vpuri3:
When you do modify `FunctionOperator`, can you propagate the kwargs down to `L.op, L.op_adjoint`?
```julia
function update_coeff...
for op in getops(L)
  if !isnothing(op)
    update_coefficients!(op, u, p, t; kw...)
  end
end
```

gaurav-arya:
check now:) 

The construction is a bit awkward, because the `FunctionOperator` already takes some `kwargs`. I did something reasonable for now, but happy to work with you on improving it in the future when #162 is addessed

test/scalar.jl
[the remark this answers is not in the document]
vpuri3: @gaurav-arya thanks for clarifying that `f(; kw)` is equivalent to `f(; kw = kw)`. In that case, discard the above comment.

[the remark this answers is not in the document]
gaurav-arya: We can indeed provide a default value to the kwarg. I could add this to the examples, although I'm not sure if I should because I don't know a well-motivated case where we would want to do this (we would need an operator that only needs partial information to updade its state)

docs/src/interface.md
gaurav-arya: Added an example!

src/func.jl
```diff
@@ -169,17 +176,17 @@ function FunctionOperator(op,
...
 
-function update_coefficients(L::FunctionOperator, u, p, t)
-    op = update_coefficients(L.op, u, p, t)
-    op_adjoint = update_coefficients(L.op_adjoint, u, p, t)
-    op_inverse = update_coefficients(L.op_inverse, u, p, t)
-    op_adjoint_inverse = update_coefficients(L.op_adjoint_inverse, u, p, t)
+function update_coefficients(L::FunctionOperator, u, p, t; kwargs...)
+    for op in getops(L)
+        op = update_coefficients(op, u, p, t; kwargs...)
+    end
```
vpuri3, line 189: since `update_coeffecients()` is oop, you gotta store the output.
```julia
new_ops = ()
for op in getops(L)
    op = update_coefficients(op, u, p, t; kwargs...)
    new_ops = (new_ops..., op)
end

L = FunctionOperator(new_ops..., L.traits, etc)
```

src/matrix.jl
```diff
@@ -105,13 +107,14 @@ an operator of size `(N, N)` where `N = size(diag, 1)` is the leading length of
...
+function DiagonalOperator(diag::AbstractVector; update_func=nothing, accepted_kwargs=())
+    _update_func = preprocess_update_func(update_func, accepted_kwargs)
+    diag_update_func = if update_func_isconstant(_update_func)
+        _update_func
     else
-        (A, u, p, t) -> (update_func(A.diag, u, p, t); A)
+        (A, u, p, t; kwargs...) -> (_update_func(A.diag, u, p, t; kwargs...); A)
     end
-    MatrixOperator(Diagonal(diag); update_func=diag_update_func)
+    MatrixOperator(Diagonal(diag); update_func=diag_update_func, accepted_kwargs=nothing)
```
vpuri3, line 117: `accepted_kwargs` is set to `()` at some places and to `nothing` at others. Can you pick one? either works.

src/func.jl
```diff
@@ -28,6 +30,7 @@ mutable struct FunctionOperator{iip,oop,T<:Number,F,Fa,Fi,Fai,Tr,P,Tt,C} <: Abst
                               traits,
                               p,
                               t,
+                              kwargs_for_op,
```
vpuri3, line 33: I think you can stick with the same name `accepted_kwargs`.

src/matrix.jl
gaurav-arya: `nothing` actually means something very different: it's an unsafe signal "don't filter this function, it's already been filtered".

I think it might be too confusing for this to be `nothing`, since in SciML that usually means that a default will be chosen for it. Would you be okay with `SciMLOperators.NoFilter()`?

src/func.jl
gaurav-arya: good point, fixed

src/func.jl
gaurav-arya: good catch, reverted to what you had originally (but with kwargs)

src/matrix.jl
vpuri3: Thanks for the clarification. `NoKwargFilter` should be good. :D

vpuri3 (approved):

gaurav-arya:
@ChrisRackauckas ready for merge. The PR body, added docs+example, and https://github.com/SciML/SciMLOperators.jl/pull/143#issuecomment-1465273444 should give the relevant context if you want to give it a look over

vpuri3 (approved):

gaurav-arya:
@ChrisRackauckas bumping this since it's a bit of a blocker for work on other operators (because the constructor style is changed slightly here), and for integration into OrdinaryDiffEq.jl.

gaurav-arya:
@vpuri3 I decided I want to finish and merge https://github.com/SciML/OrdinaryDiffEq.jl/pull/1917 before this one, if that sounds good? So I can make sure I get the design right here 🙂

gaurav-arya:
I want to revise the design `update_coefficients(u, p, t; kwargs...)` and `update_coefficients(u, p, t; extra_state=...)` where `extra_state` can be any user specified structure (e.g. a named tuple). While the former is more elegant, I think it could hurt us down the road to take up the entire kwarg space as part of the API. The latter should also reduce the implementation complexity a lot.

I'll make the change in the next day or two!

vpuri3:
I guess I'm not getting this. Can't you pass arbitrary structs into kwargs anyway? What would the `extra_state` be exactly? Here's a use case I'm playing with right now. Can you illustrate your point with this eg?

```julia
update_func(diag, u, p, t; NN = NN, x = x) = NN(x, p, st)
L = DiagonalOperator(zeros(N); update_func = update_func)

# with this branch
L(u, p, t; x = x1, NN = NN1) # == NN1(x1) .* u
L(u, p, t; x = x2, NN = NN2) # == NN2(x2) .* u
```

gaurav-arya:
The `extra_state` option would be less permissive, not more permissive.

But actually, I think I'm OK with the current approach in this PR. It's just a hard decision:) 

I am a little bit swamped with work right now but I've added you as a contributor on the branch if you'd like to resolve the conflicts before I get back to this

ChrisRackauckas (approved):
This looks reasonable enough to me.

vpuri3:
great. we just gotta clean up FunctionOperator and this this is good to go.

vpuri3:
@ChrisRackauckas this is done.

PR #149, opened 2023-02-07
check first two dimensions for `issquare`

vpuri3:
fix https://github.com/SciML/DiffEqGPU.jl/pull/229

### DIFF
--- a/src/func.jl
+++ b/src/func.jl
@@ -171,7 +171,7 @@ function FunctionOperator(op,
                     )
 end
 
-function update_coefficients(L, u, p, t)
+function update_coefficients(L::FunctionOperator, u, p, t)
     op = update_coefficients(L.op, u, p, t)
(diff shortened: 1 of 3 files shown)


vpuri3:
@ChrisRackauckas , do we need issquare to be defined on ND arrays?

vpuri3:
based on the error in DiffEqGPU, `LoadError: MethodError: no method matching isequal(::Int64, ::Int64, ::Int64)`, and the fact that the erroring test has an `EnsembleProblem`,
`@time solve(monteprob_jac, Rodas5(), EnsembleCPUArray(), dt = 0.1, trajectories = 10,
            saveat = 1.0f0) `
it looks like we should allow issquare to be true for Nd arrays if `size(A,1) == size(A,2)`.

vpuri3:
@ChrisRackauckas this is good to go

ChrisRackauckas (approved):

Issue #159, opened 2023-03-12
Overload operator application for `ComposedOperator`


vpuri3:
For `ComposedOperator` the present `update_coeffs(L, u, p, t)` interface will update each `op in L.ops` with `u`, which is not it's input.

https://github.com/SciML/SciMLOperators.jl/blob/31275d4e2b2f3671fb4c2811656dd555c4e1f24a/src/interface.jl#L22-L27

For example, with

```julia
N = 4
A = DiagonalOperator(rand(N); update_func = (d, u, p, t) -> copy!(d, u)) # A * u == u .^2
B = DiagonalOperator(rand(N); update_func = (d, u, p, t) -> copy!(d, u)) # A * u == u .^2
C = DiagonalOperator(rand(N); update_func = (d, u, p, t) -> copy!(d, u)) # A * u == u .^2

L = A ∘ B ∘ C
u = 0.1 * ones(N)

v = L(u, nothing, 0.0)
```
```julia
julia> v = L(u, nothing, 0.0)
4-element Vector{Float64}:
 0.00010000000000000003
 0.00010000000000000003
 0.00010000000000000003
 0.00010000000000000003
```
In this case, what we want to get by composing three squaring operations is `u .^ 8`, but we will get `u .^ 4` because `A,B,C` will be updated by `u, u, u`, not `u, u^2, u^4` respectively. What we want is to update `C` with `u`, `B` with its input `C * u = u .^ 2`, and `A` with `B * C * u = u .^ 4`. The operator update procedure must necessarily be interlaced with multiplication.

```julia
function (L::ComposedOperator)(u, p, t)
    v = u
    for op in reverse(L.ops)
        update_coefficients!(op, v, p, t)
        v = op * v
    end
    v
end
```
So we need to overload update-coeffs and operator application for `ComposedOperator`. It would be wise to take a second look at other composite operators to ensure such a thing isn't happening anywhere else. Maybe we need to define `update_coeffs` overload for `AdjointOperator, TransposedOperator` to pass in `u'`. Should also take a second look at FunctionOperator.

vpuri3:
@gaurav-arya , would you be able to help with this?

gaurav-arya:
Hey, I'd be happy to help, but: I don't think I agree this is a mistake? Shouldn't `u` be the state of current solution in OrdinaryDiffEq? And in general, something fixed, not changing for different operators? I don't really understand this use case.

[...]

PR #166, opened 2023-03-13
[wip] Overload operator application to correctly apply operators with nonlinear behavior

vpuri3:
Close https://github.com/SciML/SciMLOperators.jl/issues/159. Detailed discussion in issue.

### DIFF
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -538,6 +538,28 @@ end
 #Base.:*(L::ComposedOperator, u::AbstractVecOrMat) = foldl((acc, op) -> op * acc, reverse(L.ops); init=u)
 #Base.:\(L::ComposedOperator, u::AbstractVecOrMat) = foldl((acc, op) -> op \ acc, L.ops; init=u)
 
+function (L::ComposedOperator)(u, p, t)
+    v = u
+    for op in reverse(L.ops)
+        update_coefficients!(op, v, p, t)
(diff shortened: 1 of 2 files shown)


gaurav-arya:
I thought the conclusion was that this is not needed? If one really wanted to make a nonlinear op a SciMLOperator, then it should still do nothing on `update_coefficients` and do all its work in `op(u,p,t)`. This interlacing of * and update_coefficients doesn't seem right?

gaurav-arya:
And in this design, if you were to write `A * A * B`, `A` would be updated two different times with different u's. So I still maintain that u should always be the same in update_coefficients

vpuri3:
Fixed the name to reflect what is happening here. We are overloading `L(u, p, t), L(du, u, p, t)`. `update_coeffs` isn't touched.

vpuri3:
I have limited bandwidth during the week. So I'll get to this over the weekend or on some evening.

vpuri3:
TODO

- preserve order of operations in scalaroperator constructor: ie A * λ vs λ * A and apply updatecoeffs accordingly. Also need to write L(u,p,t) method for ComposedScalarOperator.
- add L(u, p, t) for InvertedOperator as well.

divital-coder:
I think this is rendered obsolete.

@ChrisRackauckas perhaps, close this as well?

PR #174, opened 2023-05-05
clean up caching for composeoperator

vpuri3:


vpuri3:
@ChrisRackauckas @xtalax this is done. can you take a look?

Issue #171, opened 2023-03-27
Implement and test OOP `update_coefficients(L,u,p,t)`, overload `Base.resize!`


vpuri3:
The latter just entails resizing function caches.

needed for https://github.com/SciML/OrdinaryDiffEq.jl/pull/1917

vpuri3:
@gaurav-arya all `Base.resize!` needs to do is resize cache vectors.

vpuri3:
@ChrisRackauckas what is correct resize behaviour for MatrixOperator?
```julia
A = rand(10,10)
L = MatrixOperator(A)
A_ = resize!(A, 5) # A[1:5,1:5]??
```

ChrisRackauckas:
MatrixOperator should probably just be an error.

Issue #165, opened 2023-03-12
Out-of-place `update_coefficients`


gaurav-arya:
There is only limited support for this function in the codebase currently. It would be interesting to see how to do this with minimal code duplication with `update_coefficients!`: if I understand @ChrisRackauckas's "don't mix mutating/not-mutating codes" philosophy correctly, we should not just fall back on a `copy` + `update_coefficients!`

vpuri3:
first, we would need to ask the user to pass a nonallocating `update_func`. Then we use `Setfield.@set!` to form a new object. There's gonna be a bunch of dup.

PR #176, opened 2023-05-07
OOP update_coefficients(...)

vpuri3:
close https://github.com/SciML/SciMLOperators.jl/issues/171, close https://github.com/SciML/SciMLOperators.jl/issues/165

### DIFF
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -232,6 +232,12 @@ end
 Base.conj(L::ScaledOperator) = conj(L.λ) * conj(L.L)
 LinearAlgebra.opnorm(L::ScaledOperator, p::Real=2) = abs(L.λ) * opnorm(L.L, p)
 
+function update_coefficients(L::ScaledOperator, u, p, t)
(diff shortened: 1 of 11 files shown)


vpuri3:
Objects like `MatrixOperator` which contain "actual data" need an in-place and an out-of-place update method. My initial thought was to pass an update function to `MatrixOperator` as in `L = MatrixOperator(A; update_func = A_update_func)` where `update_func` may have an in-place and an out-of-place method. However since the calling signature of both IIP and OOP is the same, `update_coeffs(L, u, p, t)`/ `update_coeffs!(L, u, p, t)` it would be impossible to distinguish between IIP and OOP methods.

Another solution is to ask for two separate functions: one IIP and one OOP. Eg. `MatrixOperator(A; update_func = oop_func, update_func! = iip_func)`.  This is what I am going with unless other folks have better ideas. @ChrisRackauckas  @gaurav-arya

vpuri3:
@ChrisRackauckas @gaurav-arya please review this PR. Thanks

PR #184, opened 2023-05-11
Fixes

vpuri3:
Fix a bunch of silent errors that came up due to the changes in `cache_self` method for `ComposedOperator` (in 0.2.1). I only caught the bugs when I tested https://github.com/CalculustJL/FourierSpaces.jl. I'll add FourierSpaces as a downstream test in a follow up PR.

Also made some small fixes here and there.


vpuri3:
TODO: add tests to ensure AddedOperator doesn't nest.

vpuri3:
as CI is down, i tested locally with LinearSolve, FourierSpaces to ensure nothing is breaking.

PR #195, opened 2023-06-05
fix docs homepage, add more docs

vpuri3:


vpuri3:
@ChrisRackauckas , this is good to go. After this is merged, I think we can release `1.0`

PR #200, opened 2023-06-08
[WIp] batch_dim kwarg

vpuri3:


vpuri3:
A less complicated API would be the following:

accept a kwarg `batch::Bool = false`. If `true` assume batches are in the last dimension.

```julia
    _size = if batch
        sz_in = size(input)[1:end-1] |> prod
        sz_out = size(output)[1:end-1] |> prod
        (sz_out, sz_in)
    else
        (length(output), length(input))
    end
```

ChrisRackauckas:
> accept a kwarg batch::Bool = false. If true assume batches are in the last dimension.

That makes a lot of sense.

PR #201, opened 2023-06-08
Refactor SciMLOps after #200

vpuri3:
the changes in #200 introduced `batch` kwarg to functionoperator. The API was that with `batch = true`, the last dimension of `u` will be the batch dim. With `batch = false`, the ND input array will be treated as one "column vector". This caused a lot of size mismatch errors since the rest of SciMLOperators takes the second dimension to be the batch dimension as that is what makes sense for matrices. And cases with ND arrays were handled with reshaping to 2D arrays. The principle issue became that we had different measures of length: `length(u)` for FunctionOperator with `batch = false`, `size(u)[1:end] |> prod` for `FunctionOP` with `batch = true`, and `size(u, 2)` for the rest. For an ND array and an arbitrary wrapper, it would be hard to know which measure to use. An increased reliance on `reshape` also causes confusion that may lead to unsafe behavior. Eg. applying matrices to arrays of wrong sizes should result in errors but could be done with reshape.

With this PR, I am limiting FunctionOperator w. `batch  = true` to only accept 2D arrays where the second dimension is the batch dimension. This is consistent with the rest of the package. FunctionOperator w. `batch = false` is still able to work with ND arrays as before. Further, I am removing operator evaluation and caching methods for ND arrays to avoid confusion. The only operator that is able to work with ND arrays is `FunctionOperator` w. `batch = false`. The rest of the operators work only with `AbstractVecOrMat`s where (optionally) batches are stacked in the 2nd dimension.

### new batch API for FunctionOperator
`batch` - Boolean indicating if the input/output arrays comprise of batched column-vectors stacked in a matrix. If `true`, the input/output arrays must be `AbstractVecOrMat`s, and the length of the second dimension (the batch dimension) must be the same. The batch dimension is not involved in size computation. For example, with `batch = true`, and `size(output), size(input) = (M, K), (N, K)`, the `FunctionOperator` size is set to `(M, N)`. If `batch = false`, which is the default, the `input`/`output` arrays may of any size so long as `ndims(input) == ndims(output)`, and the `size` of `FunctionOperator` is set to `(length(input), length(output))`.

- [x] do size check before mat_mul for FunctionOperator. Other concrete types, ie MatrixOp, ScalarOp manage that already.
- [x] rm `multidim.jl`
- [x] all scalar types must accept `AbstractArray`
- [x] add FunctionOp test inside a tensorprod

This is a major update. Please tag `0.3.0` after merging.


vpuri3:
looks like rm'ing the redundant functionop constructor didn't break anything. good

src/interface.jl
```diff
@@ -392,49 +348,63 @@ expmv!(v,L::AbstractSciMLOperator,u,p,t) = mul!(v,exp(L,t),u)
...
 end
 
 function Base.:(==)(L1::AbstractSciMLOperator, L2::AbstractSciMLOperator)
+    @warn """using convert-based fallback for Base.=="""
     size(L1) != size(L2) && return false
     convert(AbstractMatrix, L1) == convert(AbstractMatrix, L1)
 end
 
 Base.@propagate_inbounds function Base.getindex(L::AbstractSciMLOperator, I::Vararg{Any,N}) where {N}
+    @warn """using convert-based fallback for Base.getindex"""
```
vpuri3, line 362: added warnings to `convert`-based fallbacks.

src/matrix.jl
```diff
@@ -395,27 +395,25 @@ v = L(u, p, t) # == A * u + B * (p * t)
 ```
 
 """
-struct AffineOperator{T,AT,BT,bT,C,F,F!} <: AbstractSciMLOperator{T}
+struct AffineOperator{T,AT,BT,bT,F,F!} <: AbstractSciMLOperator{T}
```
vpuri3, line 398: rewrote operator evaluation methods for affine op to eliminate the need for cache.

src/scalar.jl
```diff
@@ -35,24 +35,24 @@ has_mul!(::AbstractSciMLScalarOperator) = true
 islinear(::AbstractSciMLScalarOperator) = true
 has_adjoint(::AbstractSciMLScalarOperator) = true
 
-Base.:*(α::AbstractSciMLScalarOperator, u::AbstractVecOrMat) = convert(Number, α) * u
```
vpuri3, line 38: since `ScalarOperator` doesn't depend on size, it can be applied to all `AbstractArray`s

src/tensor.jl
```diff
@@ -198,21 +198,21 @@ function cache_internals(L::TensorProductOperator, u::AbstractVecOrMat)
...
 
-    _ , ni = size(inner)
-    _ , no = size(outer)
+    mi , ni = size(inner)
+    _  , no = size(outer)
     k = size(u, 2)
 
     uinner = reshape(u, (ni, no*k))
-    uouter = L.cache[2]
+    uouter = reshape(L.cache[2], (no, mi * k))
```
vpuri3, line 206: array passed to `cache_operator` must be `AbstractVecOrMat` here.

src/tensor.jl
```diff
@@ -529,7 +529,9 @@ function outer_div!(L::TensorProductOperator, u::AbstractVecOrMat)
     U = reshape(U, (ni, no, k))
     C = reshape(C, (no, ni, k))
     permutedims!(C, U, PERM)
+    C = reshape(C, (no, ni * k))
     ldiv!(outer, C)
+    C = reshape(C, (no, ni, k))
     permutedims!(U, C, PERM)
```
vpuri3, line 535: added missing calls to `reshape` for converting 3D array to 2D for mul/div

src/utils.jl
```diff
@@ -1,17 +1,32 @@
 #
-function _mat_sizes(L::AbstractSciMLOperator, u::AbstractArray)
-    m, n = size(L)
-    nk = length(u)
 
-    size_in  = u isa AbstractVecOrMat ? size(u) : (n, nk ÷ n)
-    size_out = issquare(L) ? size_in : (m, size(u)[2:end]...)
+###
```
vpuri3, line 3: moved this here from interface.jl

src/utils.jl
```diff
@@ -1,17 +1,32 @@
 #
-function _mat_sizes(L::AbstractSciMLOperator, u::AbstractArray)
```
vpuri3, line 2: removed helper functions for `multidim.jl`

test/total.jl
```diff
@@ -80,17 +82,25 @@ end
...
+
     # Introduce update function for B
     B = MatrixOperator(zeros(N,N); update_func! = (A, u, p, t) -> (A .= p))
-    C = rand(N,N)
+
+    # FunctionOp
+    _C = rand(N, N) |> Symmetric
+    f(u, p, t) = _C * u
+    f(v, u, p, t) = mul!(v, _C, u)
+    C = FunctionOperator(f, zeros(N); batch = true, issymmetric = true, p = p)
```
vpuri3, line 98: test for `FunctionOp` inside a `TensorProdOp`

src/func.jl
```diff
@@ -64,28 +64,6 @@ mutable struct FunctionOperator{iip,oop,mul5,T<:Number,F,Fa,Fi,Fai,Tr,P,Tt,C} <:
     end
 end
 
-function FunctionOperator(op,
```
vpuri3, line 67: remove obsolete FunctionOp constructor with that used to flatten Nd arrays.

src/func.jl
```diff
@@ -133,7 +111,7 @@ uniform across `op`, `op_adjoint`, `op_inverse`, `op_adjoint_inverse`.
...
-* `batch` - Boolean indicating if the input/output arrays comprise of batched vectors. If `true`, the last dimension of input/output arrays is considered to be the batch dimension and is not involved in size computation. For example, let `size(output), size(input) = (M, K), (N, K)`. If `batch = true`, then the second dimension is assumed to be the batch dimension, and the `size(F::FunctionOperator) = (M, N)`. If `batch = false`, then `size(F::FunctionOperator) = (M * K, M * K)`.
+* `batch` - Boolean indicating if the input/output arrays comprise of batched column-vectors stacked in a matrix. If `true`, the input/output arrays must be `AbstractVecOrMat`s, and the length of the  second dimension (the batch dimension) must be the same. The batch dimension is not involved in size computation. For example, with `batch = true`, and `size(output), size(input) = (M, K), (N, K)`, the `FunctionOperator` size is set to `(M, N)`. If `batch = false`, which is the default, the `input`/`output` arrays may of any size so long as `ndims(input) == ndims(output)`, and the `size` of `FunctionOperator` is set to `(length(input), length(output))`.
```
vpuri3, line 114: description of change

src/func.jl
```diff
@@ -172,24 +150,56 @@ function FunctionOperator(op,
                           isposdef::Bool = false,
                          ) where{N}
 
+    # establish types
```
vpuri3, line 153: add comments, error messages to constructor

src/func.jl
```diff
@@ -316,16 +330,63 @@ function iscached(L::FunctionOperator)
     !isnothing(L.cache)
 end
 
-function cache_self(L::FunctionOperator, u::AbstractVecOrMat, v::AbstractVecOrMat)
-    !L.traits.ifcache && @debug """Cache is being allocated for a FunctionOperator
-        created with kwarg ifcache = false."""
-    @set! L.cache = zero.((u, v))
+# fix method amg bw AbstractArray, AbstractVecOrMat
```
vpuri3, line 333: cache_operator, cache_self for functionop

src/func.jl
```diff
@@ -453,63 +521,135 @@ has_mul!(::FunctionOperator{iip}) where{iip} = iip
 has_ldiv(L::FunctionOperator{iip}) where{iip} = !(L.op_inverse isa Nothing)
 has_ldiv!(L::FunctionOperator{iip}) where{iip} = iip & !(L.op_inverse isa Nothing)
 
+function _sizecheck(L::FunctionOperator, u, v)
```
vpuri3, line 524: size checks for functionop

vpuri3:
@ChrisRackauckas

src/basic.jl
```diff
@@ -8,7 +8,7 @@ struct IdentityOperator <: AbstractSciMLOperator{Bool}
 end
 
 # constructors
-IdentityOperator(u::AbstractArray) = IdentityOperator(size(u,1))
```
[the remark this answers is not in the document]
ChrisRackauckas: Why is that? IdentityOperator and NullOperator are pretty well defined on the whole cast.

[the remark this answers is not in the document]
vpuri3: I'm removing these constructors altogether. the safest thing would be to pass a length with the `NullOperator(len::Int)` constructor.

vpuri3:
@ChrisRackauckas can you take a second look sometime soon? this is holding some stuff up.

src/utils.jl
```diff
@@ -1,17 +1,32 @@
...
+    _accepted_kwargs = (accepted_kwargs === nothing) ? () : accepted_kwargs 
+    # accepted_kwargs can be passed as nothing to indicate that we should not filter 
+    # (e.g. if the function already accepts all kwargs...). 
+    return (_accepted_kwargs isa NoKwargFilter) ? _update_func : FilterKwargs(_update_func, _accepted_kwargs)
+end
+function update_func_isconstant(update_func)
+    if update_func isa FilterKwargs
+        return update_func.f == DEFAULT_UPDATE_FUNC
+    else
+        return update_func == DEFAULT_UPDATE_FUNC
```
ChrisRackauckas, line 27:
```suggestion
        return update_func.f === DEFAULT_UPDATE_FUNC
    else
        return update_func === DEFAULT_UPDATE_FUNC
```

ChrisRackauckas (approved):

Issue #196, opened 2023-06-05
Add `isconcrete` trait


vpuri3:
true if you can perform `setindex`, `getindex`, `convert(AbstractMatrix,...)` at minimal cost

PR #205, opened 2023-06-16
concretize methods

vpuri3:
fix https://github.com/SciML/SciMLOperators.jl/issues/196

@ChrisRackauckas what is the expected heuristic here? Currently this only checks if we can convert to `AbstractMatrix`. Not if that's efficient or preferable.

Does `isconcrete(L) == true` implies we can `getindex`/ `setindex!`? Do we want separate traits `can_setindex, can_getindex`?

## EDIT
We can simplify the workflow to just two functions

- `isconvertible` indicating if `L` can be converted (cheaply) to an `AbstractMatrix`. This is true for lazy combinations of ScalarOp, MatrixOp, but not for TensorProdOp, InvertedOp. For FunctionOp, the behaviour can be set with kwarg `isconvertible`.
- `concretize(L)` returns a concrete type for `L`. It is a no-op when `L` is already concrete.

workflow in downstream packages:
```julia
if need_concrete_mat
    if !isconvertible(L)
        throw(ArgumentError("method needs concrete matrix, but $L cannot be converted to a concrete type." *
                            "Please initialize $L as an `AbstractMatrix`, or `Number`"))
    end
    L = concretize(L)
end
```

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -453,15 +453,6 @@ for op in (
     end
 end
 
-for op in (
-           :sum, :prod,
-          )
-    @eval function Base.$op(L::AbstractSciMLOperator; kwargs...)
-        @warn """using convert-based fallback in $($op)."""
-        $op(convert(AbstractMatrix, L); kwargs...)
-    end
-end
-
 for pred in (
              :issymmetric,
              :ishermitian,
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -444,6 +444,15 @@ function LinearAlgebra.opnorm(L::AbstractSciMLOperator, p::Real=2)
     opnorm(concretize(L), p)
 end
 
+for op in (
+           :sum, :prod,
+          )
+    @eval function Base.$op(L::AbstractSciMLOperator; kwargs...)
+        @warn """using convert-based fallback in $($op)."""
+        $op(convert(AbstractMatrix, L); kwargs...)
+    end
+end
+
 for pred in (
              :issymmetric,
              :ishermitian,
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -294,9 +294,6 @@ concretize(L::Union{
                     AbstractMatrix,
                     Factorization,
 
-                    # Base
-                    Number,
-
                     # SciMLOperators
                     AbstractSciMLOperator,
                    }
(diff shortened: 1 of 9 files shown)


vpuri3:
I think a good set of traits would be
- `can_convert` indicating if a `convert` method is available
- `can_index` indicating if `getindex`/ `setindex` is possible

vpuri3:
@ChrisRackauckas does this match what you need for ode.jl?

vpuri3:
@ChrisRackauckas take a look?

vpuri3:
@ChrisRackauckas can you take a look? The interface is described in the PR description

vpuri3:
Question: are `convert` methods expected to return copies or the original object? i.e. Should `convert(L::MatrixOperator)` return `L.A` or `copy(L.A)`?

vpuri3:
@ChrisRackauckas ping

ChrisRackauckas:
Sorry I missed this one so much, I'm not sure how. What the reason for the preference of `concretize(L)` instead of just `convert(AbstractMatrix, L)`? The convert method would be more general and would cover `AbstractArray` types.

vpuri3:
SciMLOperators are like matrices, and ScalarOperators are like numbers. So `concretize(L)` is convenience for converting the former to AbstractMatrix, and the latter to Number. I don't see how it would be possible to convert to AbstractArray of any other shape.

ChrisRackauckas:
I see what you mean. May be hard to use in practice unless we make `concretize(L::Union{Number,AbstractArray}) = L

src/interface.jl
```diff
@@ -244,7 +260,57 @@ isconstant(::Union{
...
+"""
+concretize(L::Union{
+                    # LinearAlgebra
+                    AbstractMatrix,
+                    Factorization,
+
+                    # SciMLOperators
+                    AbstractSciMLOperator,
+                   }
+          ) =  convert(AbstractMatrix, L)
```
ChrisRackauckas, line 300:
```suggestion
                    AbstractArray,
                    Factorization,

                    # SciMLOperators
                    AbstractSciMLOperator,
                   }
          ) =  convert(AbstractArray, L)
```

ChrisRackauckas: Should that be extended to array?

vpuri3: And, in the package, I've defined `convert(::Type{AbstractMatrix}, L::MyOperator)`  methods for each operator type like

```julia
Base.convert(::Type{AbstractMatrix}, ii::IdentityOperator) = Diagonal(ones(Bool, ii.len))
```

So calling convert with AbstractArray would lead to errors:
```julia
julia> convert(AbstractArray, DiagonalOperator(ones(4)))                                                                                                                                                 [0/110]
ERROR: MethodError: Cannot `convert` an object of type 
  MatrixOperator{Float64, LinearAlgebra.Diagonal{Float64, Vector{Float64}}, SciMLOperators.FilterKwargs{typeof(SciMLOperators.DEFAULT_UPDATE_FUNC), Tuple{}}, SciMLOperators.FilterKwargs{typeof(SciMLOperators.DEFAULT_UPDATE_FUNC), Tuple{}}} to an object of type  
  AbstractArray
```

ChrisRackauckas:
Well this works for now.

gaurav-arya:
add to docs?

ChrisRackauckas:
Thanks, missed that.

PR #210, opened 2023-07-18
allow functionoperator (batch = false) to accept/return vec'd arrays

vpuri3:
@ChrisRackauckas 

with this branch,

```julia
julia> f(u, p, t) = 2u; f(v, u, p, t) = mul!(v, 2, u);                                                                                                               [30/1927]
                                           
julia> F = FunctionOperator(f, ones(4, 4));
                                           
julia> F * ones(4, 4)
4×4 Matrix{Float64}:
 2.0  2.0  2.0  2.0
 2.0  2.0  2.0  2.0
 2.0  2.0  2.0  2.0
 2.0  2.0  2.0  2.0                        

julia> F * vec(ones(4, 4))
16-element Vector{Float64}:
 2.0
 2.0
 2.0
 2.0
 2.0
...
julia> mul!(zeros(16), F, ones(16))
16-element Vector{Float64}:
 2.0
 2.0
 2.0
 2.0
 2.0
...

```

### DIFF
--- a/docs/src/tutorials/fftw.md
+++ b/docs/src/tutorials/fftw.md
@@ -19,21 +19,22 @@ x  = range(start=-L/2, stop=L/2-dx, length=n) |> Array
 u  = @. sin(5x)cos(7x);
 du = @. 5cos(5x)cos(7x) - 7sin(5x)sin(7x);
 
-k  = rfftfreq(n, 2π*n/L) |> Array
-m  = length(k)
-transform = plan_rfft(x)
+k = rfftfreq(n, 2π*n/L) |> Array
+m = length(k)
+P = plan_rfft(x)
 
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Should we just remove the `vec` in that spot of OrdinaryDiffEq if it's hitting an operator?

vpuri3:
We will remove the entire WOperator object evenetually.

The idea behind this PR is that we should support vec'd arrays anyway because one or the other downstream packages would use it. I've added warnings, docs etc all over the place as well. So after this PR, I believe we will be hitting all use cases and this package will be as versatile as it gets.

vpuri3:
We should also move preconditioner tests to interfaceII in ODE.jl

vpuri3:
BTW the failing test in `preconditioner.jl` is passing now

ChrisRackauckas (approved):

PR #213, opened 2023-07-21
Draft a ConcretizedOperator

gaurav-arya:


vpuri3:
can you explain what's going on?

gaurav-arya:
If `L` is a lazy operator, we might want to do `A = convert(AbstractMatrix, L)` because cerrtain operations might be faster with the matrix. But written like I just did, `A` loses the updating behaviour, whereas `ConcretizedOperator` does not. Similar to `_concrete_form` for `WOperator`'s.

This PR is still WIP though

PR #217, opened 2023-07-31
Only warn about convert-based fallbacks when operator not convertible.

gaurav-arya:
Gets rid of a lot of noise in OrdinaryDiffEq mass matrix tests


gaurav-arya:
ready

ChrisRackauckas (approved):

Issue #247, opened 2024-10-12
Memory allocations when solving ODE with `ScaledOperator`


albertomercurio:
**Describe the bug 🐞**

When solving a simple ODE using a `ScaledOperator`, I see many memory allocations. If I initialize the operator as a `MatrixOperator`, everything is fine. I think this issue might be related to #246. I also have the same problem using dense matrices.

**Expected behavior**

Just to have the minimum memory allocations required for an ODE. The memory allocations don't have to scale with the integration time.

**Minimal Reproducible Example 👇**

```julia
T = Float64
N = 100
A = -1 * MatrixOperator(sprand(T, N, N, 0.1))

v = rand(T, N)
A = cache_operator(A, rand(T, N))

tspan = (0.0, 2.0)

prob = ODEProblem{true}(A, v, tspan)
@benchmark solve($prob, Tsit5(), saveat=$[tspan[end]])
```

```julia
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  37.844 μs …  12.317 ms  ┊ GC (min … max): 0.00% … 99.32%
 Time  (median):     43.317 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   44.688 μs ± 122.845 μs  ┊ GC (mean ± σ):  2.74% ±  0.99%

    █▃        ▁▁▃▂                                              
  ▂▆███▇▆▆▆▆▆▆█████▇▆▅▄▄▄▃▃▃▃▃▃▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  37.8 μs         Histogram: frequency by time         64.8 μs <

 Memory estimate: 34.66 KiB, allocs estimate: 425.
```

And with a simple `MatrixOperator` everything is fine

```julia
A = MatrixOperator(-1 * sprand(T, N, N, 0.1))

v = rand(T, N)
A = cache_operator(A, rand(T, N))

tspan = (0.0, 2.0)

prob = ODEProblem{true}(A, v, tspan)
@benchmark solve($prob, Tsit5(), saveat=$[tspan[end]])
```

```julia
BenchmarkTools.Trial: 10000 samples with 1 evaluation.
 Range (min … max):  31.964 μs … 224.430 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     36.132 μs               ┊ GC (median):    0.00%
[...]

PR #249, opened 2024-10-13
Improve `mul!`, `AddedOperator`, and `update_coefficients!` to remove memory allocations

### LINKED ISSUE #246: AddedOperator of Sparse matrices allocates memory
**Describe the bug 🐞**

In my [QuantumToolbox.jl](https://github.com/qutip/QuantumToolbox.jl) package I've implemented the `OperatorSum` struct, which has the same behavior of 

$$
\sum_i \text{ScalarOperator}(c_i) * \text{MatrixOperator} (O_i)
$$

At that time, I wasn't aware of this functionality of this package. However, I saw that the `SciMLOperators` implementation allocates memory when the `MatrixOperator`s are sparse matrices. This doesn't happen in my implementation of `OperatorSum`, which should work in the same way.

**Expected behavior**

No memory allocations.

**Minimal Reproducible Example 👇**

```julia
T = Float64
N = 100
mat_list = [sprand(T, N, N, 0.1) for _ in 1:10]
coef_list = rand(T, 10)

op1 = OperatorSum(coef_list, mat_list)
op2 = mapreduce(i->ScalarOperator(coef_list[i]) * MatrixOperator(mat_list[i]), +, 1:length(coef_list))

v = rand(T, N)
v_cache1 = similar(v)
v_cache2 = similar(v)
op2 = cache_operator(op2, v)

mul!(v_cache1, op1, v);
mul!(v_cache2, op2, v);
v_cache1 ≈ v_cache2

@benchmark mul!($v_cache1, $op1, $v)

@benchmark mul!($v_cache2, $op2, $v)
```

```julia
BenchmarkTools.Trial: 10000 samples with 8 evaluations.
 Range (min … max):  4.275 μs …  7.341 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     4.415 μs              ┊ GC (median):    0.00%
 Time  (mean ± σ):   4.422 μs ± 66.127 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                     ▂▃▆▇██▇▅▃▁                               
  ▂▁▁▂▂▂▂▂▂▂▂▂▂▃▃▄▄▆███████████▇▆▅▄▃▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂▂ ▃
  4.27 μs        Histogram: frequency by time        4.61 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.

BenchmarkTools.Trial: 10000 samples with 8 evaluations.
 Range (min … max):  3.749 μs …   7.074 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     3.832 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   3.903 μs ± 186.112 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▅█▇▄                                                    
  ▁▁▂▄▇████▇▄▂▂▁▁▁▁▁▁▁▁▁▁▂▂▄▄▆▆▅▄▃▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  3.75 μs         Histogram: frequency by time        4.35 μs <

 Memory estimate: 800 bytes, allocs estimate: 20.
```

## Possible fix

In my implementation, I had similar issues, which were fixed by using the `@inline` macro in front of the `mul!` definition. In this case the macro should be defined in front of [this definition](https://github.com/SciML/SciMLOperators.jl/blob/7ea582b7554d192241196c262a65d58090a5b748/src/basic.jl#L436), but I might be wrong.

However, I have no clue why such macro would remove these allocations.

albertomercurio:
This PR fixes #246 and the remaining allocations of #247.

This is related to [this issue](https://github.com/JuliaSparse/SparseArrays.jl/issues/566) in the SparseArrays.jl package.


albertomercurio:
Ok, I'm done.

albertomercurio:
I've also noticed that the time-dependent case is not still fixed. As an example

```julia
T = ComplexF64
N = 500
M = 10
# A1 = MatrixOperator(rand(T, N, N))
# A2 = MatrixOperator(rand(T, N, N))
A1 = MatrixOperator(sprand(T, N, N, 5 / N))
A2 = MatrixOperator(sprand(T, N, N, 5 / N))
A3 = MatrixOperator(sprand(T, N, N, 5 / N))

coeff1(a, u, p, t) = sin(p.ω * t)
coeff2(a, u, p, t) = cos(p.ω * t)
coeff3(a, u, p, t) = sin(p.ω * t) * cos(p.ω * t)

c1 = ScalarOperator(rand(T), coeff1)
c2 = ScalarOperator(rand(T), coeff2)
c3 = ScalarOperator(rand(T), coeff3)

H = c1 * A1 + c2 * A2 + c3 * A3

u = rand(T, N);
du = similar(u);
p = (ω=0.1,);
t = 0.1;

@benchmark $H($du, $u, $p, $t)
```

```
BenchmarkTools.Trial: 10000 samples with 7 evaluations.
 Range (min … max):  4.964 μs …  10.530 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     5.045 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   5.079 μs ± 211.087 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▂▆▇██▇▆▅▃▂▁           ▁▁▁ ▁       ▁▁▁▁                 ▂
  ▄▁▄▃▁█████████████▆▆▅▆▆▆▆▆█████████▆▇▇██████▇▅▇▆▅▆▄▄▄▅▅▃▄▁▄ █
  4.96 μs      Histogram: log(frequency) by time      5.48 μs <

 Memory estimate: 752 bytes, allocs estimate: 8.
```

And the number of the allocations scales as the number of operators in the sum. I actually don't understand what is the problem.

Anyhow, the time-independent code in #246 is fixed.

albertomercurio:
Ok, also the time-dependent case should be fixed. At least for the case I studied. I basically followed the discussion on [this thread](https://discourse.julialang.org/t/memory-allocations-when-performing-operation-on-tuple-of-structs/121276/7), where they recommended to use generated functions when dealing with tuples of different types.

Indeed, the memory allocations are absent now.

src/basic.jl
```diff
@@ -288,13 +288,14 @@ end
...
-function LinearAlgebra.mul!(v::AbstractVecOrMat, L::ScaledOperator, u::AbstractVecOrMat)
+@inline function LinearAlgebra.mul!(
+        v::AbstractVecOrMat, L::ScaledOperator, u::AbstractVecOrMat)
     iszero(L.λ) && return lmul!(false, v)
     a = convert(Number, L.λ)
     mul!(v, L.L, u, a, false)
 end
 
-function LinearAlgebra.mul!(v::AbstractVecOrMat,
+@inline function LinearAlgebra.mul!(v::AbstractVecOrMat,
```
ChrisRackauckas, line 298: why is this needed?

src/basic.jl
```diff
@@ -441,26 +489,33 @@ function Base.:*(L::AddedOperator, u::AbstractVecOrMat)
...
         L::AddedOperator,
         u::AbstractVecOrMat,
         α,
         β)
-    lmul!(β, v)
-    for op in L.ops
-        iszero(op) && continue
-        mul!(v, op, u, α, true)
+    ops_types = L.parameters[2].parameters
+    N = length(ops_types)
```
ChrisRackauckas, line 511: is this actually required?

ChrisRackauckas:
Needs tests.

src/basic.jl
albertomercurio: It is related to [this issue](https://github.com/JuliaSparse/SparseArrays.jl/issues/566) in SparseArrays.jl. We currently need it to avoid extra allocations.

src/basic.jl
albertomercurio: The `AddedOperator` contains a `Tuple` of different objects. Performing a simple `for` loop would lead to runtime dispatch and extra allocations. Following [this thread](https://discourse.julialang.org/t/memory-allocations-when-performing-operation-on-tuple-of-structs/121276/9) they recommended to use `@generated` functions. In this way, we directly work with the single types of the `Tuple`, and we unroll the for loop.

ChrisRackauckas: A recursive implmeentation is probably cleaner and would get more compilation reuse?

albertomercurio:
I didn't add any new functionality, so the current tests should be fine. This was just to remove extra allocations and improve performance. But if you want to add some test, do you have something specific in mind?

src/basic.jl
albertomercurio: How exactly?

ChrisRackauckas:
A test for zero allocations so it doesn't regress

albertomercurio:
Ok, I added the tests. It seems that `@allocations` returns 1 instead of 0 like with `BenchmarkTools.jl`. `@allocations` returns 0 outside of the test.

Anyways, the number of allocations of the current master branch is of the order of hundreds.

ChrisRackauckas:
It's allocating a return because of the global scope IIUC. Wrap it in a function that then returns `nothing`.

albertomercurio:
Still the same. Outside the tests it returns 0 allocations with both `@allocations` and BenchmarkTools.jl. Inside the test it returns 1.

oscardssmith:
This happens because `apply_op!` is a a dispatch so it's allocating for the method call. It does seem to me that this might be a bug in the `@allocations` macro, but I'm not entirely sure. The easiest way to get around it is to something like
```
test_apply_noalloc(H, du, u, p, t) = @test @allocations apply_op!(H, du, u, p, t) ==0
```
since that will introduce a function barrier.

PR #264, opened 2025-04-07
Add special methods for multiplication of constant scalars

albertomercurio:
Here I add more methods on the multiplication of constant scalars like `Number` or `UniformScalar`.

This will simplify the structure of the nested operators, when applicable.

### DIFF
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -218,6 +218,26 @@ for T in SCALINGNUMBERTYPES
     end
 end
 
+# Special cases for constant scalars. These simplify the structure when applicable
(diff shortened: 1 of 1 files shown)


albertomercurio:
The runtests are failing but only in the Zygote implementation. It should not be related to this PR, as it is also present in other PRs.

albertomercurio:
Hello, any thoughts on this?

ChrisRackauckas (approved):

ChrisRackauckas:
Sorry I had missed this. Looks fine.

Issue #251, opened 2024-10-17
Better error report for dividing by FunctionOperator with no inverse


KevinDCarlson:
**Is your feature request related to a problem? Please describe.**

If you try to write `F \ v` for `F::FunctionOperator` with `F.op_inverse` not defined, you get the opaque error message `ERROR: MethodError: objects of type Nothing are not callable`.

**Describe the solution you’d like**

Write a clearer error message.

divital-coder:
A minimal working example to test this would be : 
```
using SciMLOperators, LinearAlgebra
using Test

# Create a simple function operator without an inverse
N = 4
f(v, u, p, t) = 2.0 * v  # A simple scaling operation
f(w, v, u, p, t) = w .= 2.0 * v  # In-place version

# Create vectors for testing
v = rand(N)
result = zeros(N)

# Create a FunctionOperator without specifying op_inverse
F = FunctionOperator(f, zeros(N), zeros(N); islinear=true)


try
    # This should fail with "objects of type Nothing are not callable"
    result = F \ v
    println("Division unexpectedly succeeded")
catch e
    println("Error type: ", typeof(e))
    println("Error message: ", e)
end

# Now try a version with op_inverse defined for comparison
finv(v, u, p, t) = 0.5 * v  # Inverse operation
finv(w, v, u, p, t) = w .= 0.5 * v

F_with_inverse = FunctionOperator(f, zeros(N), zeros(N); 
    op_inverse=finv, islinear=true)

# This should work
result = F_with_inverse \ v
println("\nDivision with proper inverse: ", result[1:min(3,length(result))], "...")
```

PR #274, opened 2025-05-18
[WIP] README update

divital-coder:
@ChrisRackauckas The operator algebra tutorial markdown file does not include correct code examples, I tried to provide working examples for operator algebra in this README.md based on new interface, but they appear to fail everytime,

Any suggestion?

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,13 @@ Let `M`, `D`, `F` be matrix-based, diagonal-matrix-based, and function-based
 
 ```julia
 N = 4
-f(u, p, t) = u .* u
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
Fixed in https://github.com/SciML/SciMLOperators.jl/pull/275

ChrisRackauckas:
Thanks for pointing it out. I realized it uncovered two potential bugs so I dove in.

ChrisRackauckas:
BTW if you have a moment, figuring out if https://github.com/SciML/SciMLOperators.jl/issues/251 got fixed would be good.

PR #289, opened 2025-07-24
Improve IdentityOperator and ScaledOperator actions

albertomercurio:
## Checklist

## Additional context

I slightly improved the `IdentityOperator` and `ScaledOperator` action. 

More specifically, I removed the `update_coefficients` in the `IdentityOperator`, as it is constant by construction.

I have also simplified the action of a `ScaledOpertator`. The previous implementation was directly using `mul!`, which I think it not always the case, as `ScaledOperator` may contain a more complex operator in `L.L`. So, I first update the coefficients of `L.λ`, and then directly apply the operator action `L.L(w, v, u, p, t)`.

After merging this PR (if accepted), could we release a new version? It is needed to add support in QuantumToolbox.jl of the new v1 version.

### DIFF
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -77,16 +77,14 @@ function (ii::IdentityOperator)(v::AbstractVecOrMat, u, p, t; kwargs...)
 end
 
 # In-place: w is destination, v is action vector, u is update vector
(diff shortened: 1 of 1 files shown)


albertomercurio:
Could you release a new version of SciMLOperators.jl and SciMLBase.jl? We need it to support SciMLOperators.jl v1 in QuantumToolbox.jl

- https://github.com/qutip/QuantumToolbox.jl/pull/470

PR #310, opened 2025-09-09
Fix type instabilities on complex AddedOperators and add new methods

albertomercurio:
## Checklist

## Additional context

The code I added in the runtests fails using the master branch. Here I have fixed some type instabilities and added a few new methods for `AddedOperator`, in order to simplify some operations.

### DIFF
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -429,7 +429,7 @@ struct AddedOperator{T,
     function AddedOperator(ops)
         @assert !isempty(ops)
         _check_AddedOperator_sizes(ops)
-        T = promote_type(eltype.(ops)...)
+        T = mapreduce(eltype, promote_type, ops)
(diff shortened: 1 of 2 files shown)


albertomercurio:
Any thoughts about it?

src/basic.jl
```diff
@@ -429,7 +429,7 @@ struct AddedOperator{T,
     function AddedOperator(ops)
         @assert !isempty(ops)
         _check_AddedOperator_sizes(ops)
-        T = promote_type(eltype.(ops)...)
+        T = mapreduce(eltype, promote_type, ops)
```
ChrisRackauckas, line 432: make sure this doesn't allocate? Usually mapreduce is expensive.

albertomercurio: I had to change it because the broadcasting was actually making type instabilities.

PR #321, opened 2025-10-28
Add operator simplification for unary negation and flatten nested AddedOperators

albertomercurio:
## Checklist

## Additional context

I add some specific methods for unary negation that may significantly simplify the expression when applicable.

Moreover, I have added a function that flattens the input `Tuple` of an `AddedOperator`. This avoids to have nested `AddedOperators` when just calling `AddedOperator(ops)`.

If this pr will be merged, could a new version be released?

### DIFF
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -243,9 +243,11 @@ end
 
 # constructors
 for T in SCALINGNUMBERTYPES[2:end]
-    @eval ScaledOperator(λ::$T, L::AbstractSciMLOperator) = ScaledOperator(
-        ScalarOperator(λ),
-        L)
+    @eval function ScaledOperator(λ::$T, L::AbstractSciMLOperator)
+        T2 = Base.promote_eltype(λ, L)
+        Λ = λ isa UniformScaling ? UniformScaling(T2(λ.λ)) : T2(λ)
+        ScaledOperator(ScalarOperator(Λ), L)
+    end
 end
 
 for T in SCALINGNUMBERTYPES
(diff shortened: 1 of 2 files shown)


src/basic.jl
```diff
@@ -286,8 +286,18 @@ for T in SCALINGNUMBERTYPES[2:end]
...
 Base.:+(L::AbstractSciMLOperator) = L
+Base.:-(L::AbstractSciMLOperator) = ScaledOperator(-true, L)
+
+# Special cases for constant scalars. These simplify the structure when applicable
+function Base.:-(L::ScaledOperator)
+    isconstant(L.λ) && return ScaledOperator(-L.λ, L.L)
+    return ScaledOperator(L.λ, -L.L) # Try to propagate the rule
+end
+function Base.:-(L::MatrixOperator)
+    isconstant(L) && return MatrixOperator(-L.A)
```
ChrisRackauckas: this allocates though

albertomercurio: Could you make an explicit example? In my mind this is my workflow

```julia
some_f(a,u,p,t) = ...

# The allocations apply here, if the matrix is constant
A = ScalarOperator(rand(), update_function=some_f) * MatrixOperator(...) + ....

prob = ODEProblem(A, u0, tspan)
sol = solve(prob, Tsit5()) # It just updates the ScalarOperators. No allocations
```

ChrisRackauckas: You'd have to do a case where someone directly writes `jac_prototype = MatrixOperator(J)` I think, then it should allocate the extra terms instead of doing the lazy computation that is expected.

PR #326, opened 2025-11-17
Avoid reset macro on AddedOperator update and caching

albertomercurio:
## Checklist

## Additional context
When differentiating an ODE with Zygote, I was having an error at the point where `cache_operator` was called. Appartently Zygote doesn't like the `@reset` macro.

See for example these errors:

https://github.com/qutip/QuantumToolbox.jl/actions/runs/19443089834/job/55630970771?pr=596

Here I avoid it using generated function. This should also make the code more robust on more structured AddedOperators.

### DIFF
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -571,13 +571,13 @@ function Base.resize!(L::AddedOperator, n::Integer)
     L
 end
 
-function update_coefficients(L::AddedOperator, u, p, t)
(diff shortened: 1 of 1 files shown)


albertomercurio:
The OrdinaryDiffEq tests should not be related to this PR. I have tested the locally with the stable version of SciMLOperators.jl and still fail

PR #332, opened 2025-12-01
Remove uses of undefined `_vec`

oscardssmith:
Because the args are AbstractVecOrMat this is equivalent to `vec` anyway.

### DIFF
--- a/src/woperator.jl
+++ b/src/woperator.jl
@@ -139,13 +139,31 @@ function Base.:\(W::WOperator, x::Number)
     end
 end
 
-function LinearAlgebra.mul!(Y::AbstractVecOrMat, W::WOperator, B::AbstractVecOrMat)
+function LinearAlgebra.mul!(Y::AbstractVector, W::WOperator, B::AbstractVector)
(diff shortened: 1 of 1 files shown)


src/woperator.jl
```diff
@@ -139,13 +139,31 @@ function Base.:\(W::WOperator, x::Number)
...
 
-function LinearAlgebra.mul!(Y::AbstractVecOrMat, W::WOperator, B::AbstractVecOrMat)
+function LinearAlgebra.mul!(Y::AbstractVector, W::WOperator, B::AbstractVector)
     # Compute mass_matrix * B
     if isa(W.mass_matrix, UniformScaling)
         a = -W.mass_matrix.λ / W.gamma
         @. Y=a*B
     else
-        mul!(_vec(Y), W.mass_matrix, vec(B))
+        mul!(Y, W.mass_matrix, B)
```
ChrisRackauckas: are you sure this actually works without this? `Y` as a matrix wouldn't work anymore, so what happens for example with `u0` a matrix?

Issue #356, opened 2026-04-21
Error when converting TensorProductOperator to sparse matrix


cvsvensson:
Converting a TensorProductOperator to a sparse matrix results in an error.

```julia
using SparseArrays, SciMLOperators
m = MatrixOperator(rand(1,1))
M = kron(m,m)
sparse(M)
#MethodError: no method matching sparse(::Type{AbstractMatrix}, ::MatrixOperator{Float64, Matrix{Float64}
```
This happens at 
https://github.com/SciML/SciMLOperators.jl/blob/4a633c7eaabb677120a3d3f5492c06f0a6b5292c/ext/SciMLOperatorsSparseArraysExt.jl#L13-L15
in SciMLOperators v1.16.0. It seems to me that the argument AbstractMatrix should not be there.

cvsvensson:
Closed by #357.

PR #357, opened 2026-04-21
fix sparse(::TensorProductOperator)

cvsvensson:
Fix the error in #356.

### DIFF
--- a/test/matrix.jl
+++ b/test/matrix.jl
@@ -1,4 +1,5 @@
 using SciMLOperators, LinearAlgebra
+using SparseArrays
 using Random
 using Test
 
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
add a test

cvsvensson:
I've added sparse(::IdentityOperator) and a test when calling sparse on a operator which consists of a combination of TensorProductOperator, AddedOperator, etc...

test/matrix.jl
```diff
@@ -655,3 +656,15 @@ end
...
+
+@testset "Sparse conversion tests" begin
+    A = rand(2, 2)
+    opA = MatrixOperator(A)
+    # Construct operator including TensorProductOperator, AddedOperator,
+    # IdentityOperator, ComposedOperator and ScaledOperator:
+    opB = kron(2 * opA + I, opA * opA)
+    B = kron(2 * A + I, A * A)
+    opB_sparse = sparse(opB)
+    @test opB_sparse ≈ B
```
ChrisRackauckas, line 668: why just approximate?

cvsvensson: I'd expect that e.g. reordering of floating point operations might lead to violations. Strict equality passed the test on my machine, but I'm not sure that's guaranteed.

This was also to match the other tests in this package, which all use approximate equalities.

PR #358, opened 2026-04-21
Simplify tensor structure and operations in the presence of IdentiyOp…

albertomercurio:
This PR simplifies the structure of a `TensorProductOperator` in presence of `IdentityOperator`. This is particularly useful when expressing operators of the form

$$
I \otimes \dots \otimes I \otimes O_i \otimes I \otimes \dots \otimes O_j \otimes I \otimes \dots \otimes I
$$

It also avoid to allocate some buffer cache arrays when the outer operator is the Identity, because the equations simplifies a lot.


ChrisRackauckas:
Add tests?

albertomercurio:
I was adding some other tests on the `mul!` function. I will open another PR then

PR #362, opened 2026-05-03
Add `mul!` tests for `TensorProductOperator`

albertomercurio:
Continuation of #358

### DIFF
--- a/src/tensor.jl
+++ b/src/tensor.jl
@@ -98,13 +98,19 @@ function TensorProductOperator(ii1::IdentityOperator, ii2::IdentityOperator)
 end
 function TensorProductOperator(ii::IdentityOperator, op::TensorProductOperator)
     left = TensorProductOperator(ii, op.ops[1])
-    # We call the main method to avoid recursion with the method below
-    return TensorProductOperator((left, op.ops[2]), nothing)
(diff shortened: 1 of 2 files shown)


src/tensor.jl
```diff
@@ -98,13 +98,19 @@ function TensorProductOperator(ii1::IdentityOperator, ii2::IdentityOperator)
 end
 function TensorProductOperator(ii::IdentityOperator, op::TensorProductOperator)
     left = TensorProductOperator(ii, op.ops[1])
-    # We call the main method to avoid recursion with the method below
-    return TensorProductOperator((left, op.ops[2]), nothing)
+    if op.ops[2] isa IdentityOperator
+        # We call the main method to avoid recursion with the method below
+        return TensorProductOperator((left, op.ops[2]), nothing)
```
ChrisRackauckas, line 103: if it's an IdentityOperator can't you just drop it?

albertomercurio: This is basically saying

$$
I \otimes (O \otimes I) = (I \otimes O) \otimes I
$$

But we need to call the main method, otherwise we call the method `TensorProductOperator(op::TensorProductOperator, ii::IdentityOperator)` below and we get `StackOverflow`.

Issue #58, opened 2022-06-17
make tensor products faster


vpuri3:
with current `@views` based implementation:

https://github.com/SciML/SciMLOperators.jl/blob/746c0d50a75067fae4121547a3c00bd4bbe27ce9/src/sciml.jl#L584-L618

```julia
using SciMLOperators, LinearAlgebra
using BenchmarkTools

A = TensorProductOperator(rand(12,12), rand(12,12), rand(12,12))

u = rand(12^3, 100)
v = rand(12^3, 100)

A = cache_operator(A, u)

mul!(v, A, u) # dunny
@btime mul!($v, $A, $u); # 16.901 ms (348801 allocations: 45.51 MiB)

B = convert(AbstractMatrix, A)
mul!(v, B, u) # dummy
@btime mul!($v, $B, $u); # 10.337 ms (0 allocations: 0 bytes)
```

```julia
julia> versioninfo()
Julia Version 1.8.0-rc1
Commit 6368fdc6565 (2022-05-27 18:33 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: 4 × Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, broadwell)
  Threads: 4 on 4 virtual cores
Environment:
  JULIA_NUM_PRECOMPILE_TASKS = 4
  JULIA_DEPOT_PATH = /Users/vp/.julia
  JULIA_NUM_THREADS = 4
```

vpuri3:
```julia
using SciMLOperators, LinearAlgebra
using BenchmarkTools

A = TensorProductOperator(rand(12,12), rand(12,12), rand(12,12))

u = [rand(12^3) for i=1:100]
v = [rand(12^3) for i=1:100]

A = cache_operator(A, u[1])

mul!(v[1], A, u[1]) # dunny
@btime for i=1:length(u); mul!($v[i], $A, $u[i]); end; # 7.333 ms (3301 allocations: 27.05 MiB)
```

so applying the operator to individual vectors is indeed faster. likely because we are avoiding the `views` call

vpuri3:
@chriselrod  @ChrisRackauckas can you take a look please?

vpuri3:
i think the 'slow' part is that we are applying the operator to each column vector in `u` one at a time
[...]

PR #369, opened 2026-05-05
Add LoopVectorization tensor-product fast path extension

AshtonSBradley:
## Revision note

Thanks for the early review comments. I revised the PR to separate fast-path availability/control flow from the mutating implementation call, rewrote the `k == 1`/batched fast-path branch so the flow is explicit, and removed unrelated cache-slice edits plus a redundant test block. Hopefully this is now back on track as a minimal extension-backed optimization.

## Summary

This PR adds an optional fast path for batched `TensorProductOperator` multiplication when the outer operator is a `MatrixOperator` wrapping a CPU `StridedMatrix`.

The core tensor implementation now exposes a small internal backend-neutral hook, `_tensor_outer_mul_fast!`, guarded by a separate `_has_tensor_outer_mul_fast` predicate. The existing generic implementation remains the fallback and still uses only the operator interface (`mul!`) after reshaping/permuting into the layout it needs. That keeps arbitrary matrix-free operators, sparse matrices, GPU arrays, and other non-strided operators on the abstraction-preserving path.

A new `SciMLOperatorsLoopVectorizationExt` extension overloads this hook with an `@turbo` loop for the narrow dense-strided `MatrixOperator` case. `LoopVectorization` is a weak dependency and a test dependency, not a hard dependency of SciMLOperators.

## Backend design

The hook is intentionally not named after LoopVectorization. Core owns a generic availability predicate plus an implementation hook:

```julia
_has_tensor_outer_mul_fast(outer) = false
function _tensor_outer_mul_fast! end
```

The predicate keeps control flow separate from the mutating fast-path call. Backend extensions opt in by overloading the predicate for supported operator types and providing the corresponding `_tensor_outer_mul_fast!` methods.

That means LoopVectorization is only the first backend. If LoopModels, or another successor backend, becomes the preferred implementation later, it can live in its own extension and overload the same hook without changing the core tensor-product algorithm or public API.

The intended backend-switching model is:

- Core SciMLOperators owns the generic fallback and the backend-neutral hook.
- `SciMLOperatorsLoopVectorizationExt` owns today’s `@turbo` dense-strided implementation.
- The same extension design could support an opt-in `@tturbo` method later, for users who want to experiment with threaded LoopVectorization behavior, without changing the core tensor code. This PR intentionally uses only `@turbo` as the safer default because it avoids thread oversubscription concerns with threaded BLAS.
- A future `SciMLOperatorsLoopModelsExt` can provide the same hook for LoopModels once its API is ready.
- A future GPU backend could also overload the same hook from a CUDA/KernelAbstractions-style extension, but this PR does not add a GPU kernel.
- If no backend is loaded, behavior remains unchanged.

## GPU compatibility

This PR should not change existing GPU behavior by dispatch. The new LoopVectorization method only applies to `MatrixOperator` wrapping a `StridedMatrix`, so GPU arrays such as `CuArray` should not be caught by the CPU scalar-indexing `@turbo` path. GPU-backed operators continue to use the existing generic fallback, which relies on the current `permutedims!`/`mul!` behavior for those array/operator types.

I did not run GPU tests locally, so this is a dispatch/abstraction argument rather than a measured GPU validation.

## Performance notes

Local benchmarks on the existing `benchmarks/tensor.jl` setup showed the extension active only when `LoopVectorization` is loaded. These numbers were collected on Apple Silicon, so they should be treated as a conservative local check rather than the expected best case. The improvement may be larger on Intel AVX/AVX2 machines, which are closer to the hardware context discussed in #58 and where LoopVectorization's SIMD code generation has historically shown larger gains.

I also tried the other idea from #58: replacing explicit `permutedims!` calls with `PermutedDimsArray`. That benchmarked substantially worse for the dense batched cases, likely because BLAS then sees a lazy/strided permuted input instead of contiguous storage. This PR therefore keeps the existing explicit permutation fallback and only bypasses it for the narrow `@turbo` extension path.

Without `LoopVectorization`, the fallback path remains at baseline:

```text
⊗(A, B) cached mul!:          ~29.1 μs, 5 allocs
⊗(A, I) cached mul!:          ~19.8 μs, 4 allocs
⊗(⊗(A, B), C) cached mul!:   ~283.7 μs, 10 allocs
⊗(A, ⊗(B, C)) cached mul!:   ~283.5 μs, 10 allocs
```

With `LoopVectorization` loaded:

```text
⊗(A, B) cached mul!:          ~17.0 μs, 3 allocs
⊗(A, I) cached mul!:           ~7.1 μs, 2 allocs
⊗(⊗(A, B), C) cached mul!:   ~234.5 μs, 8 allocs
⊗(A, ⊗(B, C)) cached mul!:   ~188.7 μs, 6 allocs
```

This follows the discussion in #58: avoid the generic `permutedims! + mul! + permutedims!` path only where fast indexing is actually valid, rather than assuming all operators are indexable.

## Tests

- Added `LoopVectorization` to the test environment so the existing dense batched `TensorProductOperator` plain/scaled `mul!` tests exercise the extension path.
- `Pkg.test()` passes locally with `LoopVectorization` in the test environment:

```text
SciMLOperators | 823 passed, 2 broken
```

### DIFF
--- a/src/tensor.jl
+++ b/src/tensor.jl
@@ -413,6 +413,9 @@ end
 # helper functions
 const PERM = (2, 1, 3)
 
+_has_tensor_outer_mul_fast(outer) = false
+function _tensor_outer_mul_fast! end
+
 function outer_mul(L::TensorProductOperator, v::AbstractVecOrMat, C::AbstractVecOrMat)
     outer, inner = L.ops
 
@@ -465,6 +468,11 @@ function outer_mul!(w::AbstractVecOrMat, L::TensorProductOperator, v::AbstractVe
         return w
     end
 
+    if _has_tensor_outer_mul_fast(outer)
+        _tensor_outer_mul_fast!(w, outer, C1, mi, mo, no, k)
+        return w
+    end
+
     C2, C3 = L.cache[2:3]
 
     C1 = reshape(C1, (mi, no, k))
@@ -500,6 +508,11 @@ function outer_mul!(
         W = reshape(w, (mi, mo))
         C = reshape(v, (mi, no))
         mul!(transpose(W), outer, transpose(C), α, β)
+        return w
+    end
+
+    if _has_tensor_outer_mul_fast(outer)
+        _tensor_outer_mul_fast!(w, outer, v, mi, mo, no, k, α, β)
         return w
     end
 
--- a/Project.toml
+++ b/Project.toml
@@ -10,10 +10,12 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Accessors = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 
 [weakdeps]
+LoopVectorization = "bdcacae8-1622-11e9-2a5c-532679323890"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 
 [extensions]
+SciMLOperatorsLoopVectorizationExt = "LoopVectorization"
 SciMLOperatorsSparseArraysExt = "SparseArrays"
 SciMLOperatorsStaticArraysCoreExt = "StaticArraysCore"
 
@@ -22,6 +24,7 @@ Accessors = "0.1.42"
 ArrayInterface = "7.19"
 DocStringExtensions = "0.9.4"
 LinearAlgebra = "1.10"
+LoopVectorization = "0.12"
 SparseArrays = "1.10"
 StaticArraysCore = "1"
 julia = "1.10"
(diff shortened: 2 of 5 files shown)


src/tensor.jl
```diff
@@ -546,6 +413,11 @@ end
 # helper functions
 const PERM = (2, 1, 3)
 
+_tensor_outer_mul_fast!(w, outer, C, mi::Int, mo::Int, no::Int, k::Int) = false
+function _tensor_outer_mul_fast!(w, outer, C, mi::Int, mo::Int, no::Int, k::Int, α, β)
+    return false
+end
+
```
ChrisRackauckas: seems weird to mix the control flow with the actual call?

src/tensor.jl
```diff
@@ -596,9 +466,13 @@ function outer_mul!(w::AbstractVecOrMat, L::TensorProductOperator, v::AbstractVe
         C1 = reshape(C1, (mi, no))
         mul!(transpose(W), outer, transpose(C1))
         return w
+    elseif _has_tensor_outer_mul_fast(outer)
```
ChrisRackauckas: if it's the elseif, isn't it bypassed?

src/tensor.jl
AshtonSBradley: this has been separated

src/tensor.jl
AshtonSBradley: yes, good catch

PR #370, opened 2026-05-06
Share same cache among similar operators in `AddedOperator`

albertomercurio:
## Checklist

## Additional context

When caching an `AddedOperator`, we currently cache all the suboperators iteratively. However, this is suboptimal, as many of them can in principle share the same cache when possible.

In this PR I implement this cache sharing in the `AddedOperator`. Let's imagine I have the sum of `Composed + Tensor + Tensor + Composed + Matrix + Tensor`, many of them require some cache, but they could share the same cache. We can be sure they can share the same cache if

1. They have the same concrete type (e.g., two `ComposedOperator`s built from the same operand types)
2. The size of their cache is the same

The first is checked at compile time, and it also pins the element types for free. The second is checked at run time by comparing the operator's size and, recursively, its operands' sizes. Nothing has to declare its cache layout: `cache_self` stays the only description of what a cache is, so there is no second one that can drift out of sync with it.

## Benchmarks

This method allows to save a lot of memory, as also shown in this example

```julia
using SparseArrays
using SciMLOperators
using CairoMakie

function generate_op(N)
    M = 10
    sparsity = 1 / N

    A_list = ntuple(i -> MatrixOperator(sprand(N, N, sparsity)) * MatrixOperator(sprand(N, N, sparsity)), Val(M))

    op = SciMLOperators.AddedOperator(A_list)

    u = rand(N, N)
    return cache_operator(op, u)
    # return op
end

# %%

N_list = 4 .^ (2:7)
sizes_main = [Base.summarysize(generate_op(N)) for N in N_list]
sizes_pr = [Base.summarysize(generate_op(N)) for N in N_list]

# %%

fig = Figure()
ax = Axis(fig[1, 1], xlabel="N", ylabel="Size (MB)", xscale=log10, yscale=log10)

scatterlines!(ax, N_list, sizes_main ./ 1e6, label="main")
scatterlines!(ax, N_list, sizes_pr ./ 1e6, label="PR")

axislegend(ax; position = :lt)

fig
```

image

**Defining a new operator with a cache**

Nothing changes unless you opt in:

```julia
# take part in sharing
SciMLOperators.getcache(op::MyOperator) = op.cache

# optional: also skip allocating a cache that would be immediately discarded
SciMLOperators.adopt_cache(op::MyOperator, cache, v) =
    cache_internals(update_cache(op, cache), v)
```

The second line is for operators following the `cache_self`/`cache_internals` split; one that defines its own `cache_operator` should instead do whatever that does, with `cache` in place of the allocation. Both default to declining, so an operator defining neither behaves exactly as before.

### DIFF
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -42,6 +42,20 @@ SciMLOperators.has_tensor_outer_mul_fast
 SciMLOperators.tensor_outer_mul_fast!
 ```
 
+### Sharing scratch space between the summands of an `AddedOperator`
+
+`mul!` applies the summands of an `AddedOperator` one after another, so no two of their
+caches are ever live at the same time and a summand can reuse the scratch of an earlier
+one. An operator type takes part by defining `getcache`, and can additionally avoid
+allocating a cache it would only discard by defining `adopt_cache`. Both default to
+declining, so a type that defines neither behaves exactly as it did before.
+
+```@docs
+SciMLOperators.getcache
+SciMLOperators.update_cache
+SciMLOperators.adopt_cache
+```
+
 ## Note About Affine Operators
 
 Affine operators are operators that have the action `Q*x = A*x + b`. These operators have
--- a/docs/src/interface.md
+++ b/docs/src/interface.md
@@ -50,6 +50,17 @@ one. An operator type takes part by defining `getcache`, and can additionally av
 allocating a cache it would only discard by defining `adopt_cache`. Both default to
 declining, so a type that defines neither behaves exactly as it did before.
 
+A wrapper holding no scratch of its own, such as `ScaledOperator`, forwards all three hooks
+to the operator it wraps, so `2A + 3B` and `A - B` share as `A + B` does. A new wrapper of
+that kind should do the same.
+
+!!! warning "Concurrency"
+    
+    Applying a cached operator from several tasks at once was never safe, because they
+    would write to the same scratch. Once its summands share scratch they are not
+    independently safe either, so code that hand-parallelizes over the `ops` of an
+    `AddedOperator` needs its own uncached copies.
+
 ```@docs
 SciMLOperators.getcache
 SciMLOperators.update_cache
--- a/src/SciMLOperators.jl
+++ b/src/SciMLOperators.jl
@@ -334,7 +334,8 @@ export update_coefficients!,
             :ScaledOperator, :AddedOperator, :ComposedOperator,
             :InvertedOperator, :AdjointOperator, :TransposedOperator,
             :AddedScalarOperator, :ComposedScalarOperator, :InvertedScalarOperator,
-            :has_tensor_outer_mul_fast, :tensor_outer_mul_fast!
+            :has_tensor_outer_mul_fast, :tensor_outer_mul_fast!,
+            :getcache, :update_cache, :adopt_cache
         )
     )
 end
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -477,6 +477,22 @@ function cache_internals(L::ScaledOperator, v::AbstractVecOrMat)
     return L
 end
 
+# A scaling holds no scratch of its own, so it is transparent to the sharing described in
+# `getcache`: the cache one field down is exactly as interchangeable as one held directly.
+# Forwarding matters more than it looks — `A - B` lowers to `AddedOperator(A, -B)` with
+# `-B` a `ScaledOperator`, so without this the commonest sum of all shares nothing.
+getcache(L::ScaledOperator) = getcache(L.L)
+update_cache(L::ScaledOperator, new_cache) = @reset L.L = update_cache(L.L, new_cache)
+
+function adopt_cache(L::ScaledOperator, cache, v)
+    inner = adopt_cache(L.L, cache, v)
+    inner === nothing && return nothing
+
+    @reset L.L = inner
+    @reset L.λ = cache_operator(L.λ, v)
+    return L
+end
+
 # getindex
 Base.getindex(L::ScaledOperator, i::Int) = L.coeff * L.L[i]
 Base.getindex(L::ScaledOperator, I::Vararg{Int, N}) where {N} = L.λ * L.L[I...]
--- a/src/basic.jl
+++ b/src/basic.jl
@@ -789,8 +789,23 @@ has_adjoint(L::AddedOperator) = all(has_adjoint, L.ops)
 @generated function cache_internals(L::AddedOperator, v::AbstractVecOrMat)
     ops_types = L.parameters[2].parameters
     N = length(ops_types)
-    cached = [:(cache_operator(L.ops[$i], v)) for i in 1:N]
-    return :(AddedOperator($(Expr(:tuple, cached...))))
+    syms = [Symbol(:op_, i) for i in 1:N]
+
+    # `mul!` applies the summands one after another, so no two of their caches are ever
+    # live at the same time and a summand can reuse the scratch of an earlier one. Only
+    # summands of the same concrete type can qualify — that is what pins their element
+    # types — so the candidates are picked out here, at compile time, and each summand
+    # tries them all before allocating anything of its own.
+    stmts = map(1:N) do i
+        donors = Expr(:tuple, syms[findall(j -> ops_types[j] === ops_types[i], 1:(i - 1))]...)
+        earlier = Expr(:tuple, syms[1:(i - 1)]...)
+        return :($(syms[i]) = _cache_summand(L.ops[$i], $donors, $earlier, v))
+    end
+
+    return quote
+        $(stmts...)
+        return AddedOperator($(Expr(:tuple, syms...)))
+    end
 end
 
 getindex(L::AddedOperator, i::Int) = sum(op -> op[i], L.ops)
@@ -1003,6 +1018,8 @@ function update_coefficients(L::ComposedOperator, u, p, t; kwargs...)
 end
 
 getops(L::ComposedOperator) = L.ops
+getcache(L::ComposedOperator) = L.cache
+adopt_cache(L::ComposedOperator, cache, v) = cache_internals(update_cache(L, cache), v)
 
 # Copy method to avoid aliasing
 function Base.copy(L::ComposedOperator)
@@ -1345,6 +1362,8 @@ function update_coefficients(L::InvertedOperator, u, p, t; kwargs...)
 end
 
 getops(L::InvertedOperator) = (L.L,)
+getcache(L::InvertedOperator) = L.cache
+adopt_cache(L::InvertedOperator, cache, v) = cache_internals(update_cache(L, cache), v)
 islinear(L::InvertedOperator) = islinear(L.L)
 isconvertible(::InvertedOperator) = false
 
--- a/src/func.jl
+++ b/src/func.jl
@@ -212,6 +212,17 @@ function set_cache(
     )
 end
 
+getcache(L::FunctionOperator) = L.cache
+# `cache` is a type parameter here, so it cannot be swapped with `@reset`.
+update_cache(L::FunctionOperator, new_cache) = set_cache(L, new_cache)
+
+# Deliberately no `adopt_cache` method: `cache_self` sizes its buffers from
+# `traits.sizes`, the input/output *array* shapes, which `size(L) == traits.size` does not
+# expose, so two same-typed operators of equal `size` can still need different buffers.
+# `cache_operator` also does batch-size fixup that skipping straight to `cache_internals`
+# would miss. `FunctionOperator` still shares via the post-hoc path in `_deduplicate_cache`,
+# which compares buffers that have actually been allocated.
+
 function input_eltype(
         ::FunctionOperator{
             iip, oop, mul5, T, F, Fa, Fi, Fai, Tr, U, P, Tt, C,
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -193,6 +193,20 @@ earlier one whose cache has the identical type and identically sized slots, so o
 this for operators where two such caches really are interchangeable — including which
 slots deliberately alias each other. The `nothing` default keeps a type out of it entirely.
 
+Two further conditions come with defining it. The first is on the operator's own
+`cache_self`: it must treat the buffers it is handed as a template to `similar`, `zero` or
+`reshape` and never retain a reference to them, since the whole point is that those buffers
+may afterwards be replaced by an earlier summand's. Every `cache_self` in this package does.
+The second is on the caller: applying the summands of an `AddedOperator` concurrently, by
+hand, is no longer safe once they share scratch — a cached operator was never safe to apply
+concurrently, but now its summands are not independently safe either.
+
+A wrapper that holds no scratch of its own — `ScaledOperator`, `AdjointOperator`,
+`TransposedOperator` — is transparent to all of this and forwards `getcache`,
+[`update_cache`](@ref) and [`adopt_cache`](@ref) to the operator it wraps. A new wrapper of
+that kind should do the same; without it the operator underneath drops out of sharing even
+though its cache would have qualified.
+
 See also [`adopt_cache`](@ref), which additionally spares the operator from allocating a
 cache it would only discard.
 """
@@ -210,8 +224,22 @@ update_cache(op::AbstractSciMLOperator, new_cache) = @reset op.cache = new_cache
 # are left to check. Anything else is conservatively treated as not interchangeable.
 _slots_match(a::AbstractArray, b::AbstractArray) = size(a) == size(b)
 _slots_match(::Nothing, ::Nothing) = true
-_slots_match(a::Tuple, b::Tuple) = all(map(_slots_match, a, b))
+function _slots_match(a::NTuple{N, Any}, b::NTuple{N, Any}) where {N}
+    return all(map(_slots_match, a, b)) && _aliasing_matches(a, b)
+end
 _slots_match(a, b) = false
+
+# Which slots deliberately point at the same buffer is part of the layout, not an accident
+# of it: `TensorProductOperator` folds two of its slots into one exactly when both factors
+# are square. Comparing the pattern keeps an operator that needs distinct buffers from ever
+# being handed a cache that shares one, instead of relying on the sizes to give it away.
+function _aliasing_matches(a::NTuple{N, Any}, b::NTuple{N, Any}) where {N}
+    return all(
+        ntuple(Val(N)) do i
+            all(ntuple(j -> (a[i] === a[j]) === (b[i] === b[j]), Val(N)))
+        end
+    )
+end
 
 # Having the same cache type is settled at compile time; only the sizes cost anything at
 # run time. Caches of differing type take the second method and are never interchanged.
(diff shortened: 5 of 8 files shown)


albertomercurio:
I think this PR is ready. It improves the cache handling for `AddedOperator`

ChrisRackauckas-Claude (changes requested):
Thanks for the work on this — the memory wins are real for the benchmark case, but I think there are correctness issues that need to be fixed before merging. I dug into the diff, ran reproducers locally against this branch on Julia 1.11.9, and found the following.

## 1. `InexactError` on mixed-eltype `ComposedOperator`s in the same `AddedOperator` (correctness)

A real-world QM-style pattern fails:

```julia
using SciMLOperators, LinearAlgebra
Ar = MatrixOperator(rand(Float64, 4, 4));    Br = MatrixOperator(rand(Float64, 4, 4))
Ac = MatrixOperator(rand(ComplexF64, 4, 4)); Bc = MatrixOperator(rand(ComplexF64, 4, 4))
L  = (Ar*Br) + (Ac*Bc)
Ls = cache_operator(L, rand(Float64, 4))
mul!(similar(rand(Float64,4), ComplexF64), Ls, rand(Float64,4))
# ERROR: InexactError: Float64(... + 0.121im)
```

Both sub-ops are `ComposedOperator`, so the asker `Ac*Bc` is forced to share `Ar*Br`'s cache. The donor's cache is `Float64` (because `v` is real and `cache_self` uses `promote_type(eltype(op), eltype(zero(v)))`). The asker would normally allocate `ComplexF64` slots, since its inner `Bc` is `ComplexF64`. But `_cache_compatible` at `src/interface.jl:210` only checks `size(hint) == shape` — **no eltype check** — so sharing proceeds. Then `mul!(L.cache[1], Bc, v)` writes a complex result into a `Float64` buffer.

The reverse order `(Ac*Bc) + (Ar*Br)` also breaks, in the 5-arg path: `copy!(cache, w)` with `w::ComplexF64` and `cache::Float64`.

Worth noting that `Ls * v` happens to return correct results because `Base.:*(L::AddedOperator, v)` (`src/basic.jl:676`) does an out-of-place `sum(op -> op*v, L.ops)` that bypasses the cache entirely. So the new testset asserts `L * u ≈ expected`, but that doesn't exercise the in-place path that the cache is built for. The `mul!(w, L, v)` path is what users on hot loops will hit.

## 2. `_get_cache_shapes(::ComposedOperator)` is off-by-one (logic bug)

```julia
# A is 5×3, B is 3×4, C = A*B, v: 4-vec
SciMLOperators._get_cache_shapes(cache_operator(A*B, v), v)
# ((5,), (3,))         ← claimed
map(size, cache_operator(A*B, v).cache)
# ((3,), (4,))         ← actual
```

`_get_cache_shapes` at `src/basic.jl:959` claims slot `i` has shape `(size(L.ops[i],1), K)`. But `cache_self` at `src/basic.jl:969` actually allocates slot `i` (for `i<N`) with the *next* operator's row dim — `(size(L.ops[i+1],1), K)` — and slot `N` with `size(v)`, not `(size(L.ops[N],1), K)`.

Effect:
- For non-square `ComposedOperator`s, the spec-vs-actual mismatch makes `_cache_compatible` reject the donor's cache → sharing never triggers for any non-square inner ops. The benchmark wins disappear in this regime.
- For square inner ops, the bug is masked because `size(op,1) == size(op,2)`. But that's also exactly where the eltype-mismatch bug bites (#1).

Better to have a single source of truth here — either compute shapes inside `cache_self` and stash them, or factor out a helper that both functions use, so they can't drift.

## 3. `FunctionOperator._get_cache_shapes` ignores `size(v,2)` and eltype

`src/func.jl:593`:

```julia
function _get_cache_shapes(L::FunctionOperator, v::AbstractVecOrMat)
    return (L.traits.sizes[1], L.traits.sizes[2])
end
```

For matrix input it returns single-column shapes. That'll always reject sharing for a matrix `v`, but could also falsely match against a same-vector-shaped donor and graft the wrong buffer. Should incorporate `size(v,2)` and `output_eltype`.

## 4. No device check in `_cache_compatible`

A CPU donor and GPU asker (or vice versa) with the same `size(...)` would silently graft a wrong-device cache. Unlikely to occur in normal usage but the API permits it.

## 5. Aliasing is brittle — worth a comment

After sharing, several `L.ops[i].cache` are `===`. This works only because `AddedOperator`'s `mul!` is strictly serial. Any future change that parallelizes `L.ops` evaluation (Threads, separate CUDA streams, async) creates data races on the shared buffers. A comment near the donor selection or near the cache field would help future maintainers not break this invariant.

## 6. Test coverage gap

The new testset at `test/basic.jl:411` only uses square `Float64` `MatrixOperator(rand(N,N))` matrices and asserts `L * u` (the path that bypasses the cache). None of the failure modes above are covered. Suggest adding:
- Mixed-eltype `ComposedOperator`s (e.g. real + complex inner ops).
- Non-square inner ops in `ComposedOperator`.
- `FunctionOperator` inside an `AddedOperator` with both vector and matrix `v`.
- `mul!(w, L, v)` and `mul!(w, L, v, α, β)` against the cached `AddedOperator` (not just `*`).
- The benchmark in the PR description also only measures `Base.summarysize`; an in-place `mul!` benchmark would catch correctness regressions of this kind.

## Suggested minimal fixes

1. Extend `_cache_compatible` (or `_get_cache_shapes`) to carry and check eltype.
2. Fix `_get_cache_shapes(::ComposedOperator)` so it matches what `cache_self` allocates, ideally by sharing a common helper.
3. Fix `_get_cache_shapes(::FunctionOperator)` to include `size(v,2)` and `output_eltype`.
4. Add the missing tests above so the eltype/non-square/in-place paths are covered.

Happy to take a closer look at a follow-up commit if useful. Reproducers I used: <details><summary>repro scripts</summary>

```julia
# repro3.jl — confirms #1
using SciMLOperators, LinearAlgebra
Ar = MatrixOperator(rand(Float64, 4, 4));    Br = MatrixOperator(rand(Float64, 4, 4))
Ac = MatrixOperator(rand(ComplexF64, 4, 4)); Bc = MatrixOperator(rand(ComplexF64, 4, 4))
CR = Ar * Br; CC = Ac * Bc
L = CR + CC
v = rand(Float64, 4)
Ls = cache_operator(L, v)
@show map(eltype, Ls.ops[1].cache) map(eltype, Ls.ops[2].cache) Ls.ops[2].cache === Ls.ops[1].cache
w = similar(v, ComplexF64); fill!(w, 0)
mul!(w, Ls, v)   # InexactError
```

```julia
# repro.jl — confirms #2
using SciMLOperators
A = MatrixOperator(rand(5, 3)); B = MatrixOperator(rand(3, 4))
C = A * B; v = rand(4)
Cc = cache_operator(C, v)
@show SciMLOperators._get_cache_shapes(Cc, v)   # ((5,), (3,))
@show map(size, Cc.cache)                       # ((3,), (4,))
```

</details>

albertomercurio:
Ok I should have fixed all the issues. I'm not sure about the FunctionOperator. But the others should be fixed

albertomercurio:
Any suggestions about this?

albertomercurio:
Hello, are there any news about this?

albertomercurio:
Hello, this has been stuck for a while. Is there a reason? I believe that this will improve the memory efficiency of such cases.

I am open to other critiques but I would like to have this feature

oameye:
@albertomercurio We just have to keep pinging :)

ChrisRackauckas:
Is this safe though? Checking its safety in all cases is really hard.

albertomercurio:
By default ha new `AbstractSciMLOperator` has the property to not share the cache, since `getcache(::AbstractSciMLOperator) = nothing`. It is only defined for ComposedOperator, InvertedOperator and FunctionOperator. So, any new operator definition would work as before, that is not sharing the cache, if not definition is given.

albertomercurio:
@ChrisRackauckas I have totally changed the implementation. It should be better than the previous one. It is a safe implementation in my opinion, as no breaking changes are introduced.

**What it does now.** Summands are grouped by *concrete type*, which is settled at compile time and pins every element type for free. The only run-time check is that their operand sizes agree. A summand then takes an earlier one's buffers directly, before allocating anything of its own.

Why this is better:

- **Nothing is declared.** `cache_self` stays the single description of what a cache is, so there is nothing that can drift out of sync with it. `_get_cache_shapes` and `_cache_compatible` are gone, and `TensorProductOperator.cache_self` is back to untouched.
- **Element-type safety is free** instead of something to get right — it falls out of grouping by concrete type rather than by wrapper, which is what the bug above came from.
- **Opt-in, declining by default.** A type shares only if it defines `getcache`, and skips allocation only if it defines `adopt_cache`. Nothing is exported.
- **Smaller:** +217/−2, down from +252/−19, with no existing cache-construction code touched.

ChrisRackauckas-Claude:
## Verdict

I went through this carefully and could not find a case where the sharing produces a wrong answer. The soundness argument holds. My concerns are all about *generality* — the rule that decides who gets to share is narrower and more structure-dependent than the description suggests, and it silently produces nothing in several of the shapes an `AddedOperator` most often arrives in.

## What I ran

On the PR branch (merge base = current master `1b29c59`), Julia 1.11.9:

| Check | Result |
|---|---|
| `GROUP=Core` | passed |
| `GROUP=QA` (Aqua / ExplicitImports) | passed |
| `GROUP=Downstream` (alloc checks) | passed |
| 49 adversarial assertions I wrote | all passed |

Adversarial cases: same-typed composed summands with *different* inner dims in one sum; tensor products with the same total size but a different factor split (`2⊗4` vs `4⊗2`); non-square tensor factors; matrix input (`K > 1`); `InvertedOperator` summands through 5-arg `mul!` (which parks `w` in the cache); the same operator object appearing twice in the sum; mixed complex/real cache eltypes; `update_coefficients!` followed by `mul!`; nested composed-of-composed; `adjoint`; `copy`; and repeated/interleaved `mul!` calls checked against dense references. `@inferred cache_operator` is clean on all of them, including heterogeneous sums.

The new allocation test (`setup(8) < 2*setup(2)`) is not flaky — 5 seeds, consistent 2× margin.

## Why it is sound

The load-bearing claim — `AddedOperator` applies summands strictly serially — is true on all four application paths (`src/basic.jl:800-846`), and the out-of-place path never touches caches. Two independent guards back it: the pre-emptive `adopt_cache` path requires identical *concrete summand type* (which pins the cache type and every eltype) plus a recursive shape signature; the post-hoc `_deduplicate_cache` path compares buffers that already exist, by type and size, so it can't drift out of sync with `cache_self`. Only four types opt in and all four caches are pure scratch. Downstream operators with a `cache` field get `getcache -> nothing` and are unaffected, so it is backward compatible.

One invariant is load-bearing but unstated, and I'd add it to the `getcache` docstring: **the post-hoc swap is safe only because no `cache_self`/`cache_operator` in the package retains a reference to the array it is handed.** They all use it purely as a `similar`/`zero`/`reshape` template — I checked every one (`ComposedOperator`, `TensorProductOperator`, `InvertedOperator`, `FunctionOperator`, `AffineOperator`, `InvertibleOperator`, `BlockDiagonalOperator`, the `left.jl` wrappers). If a future `cache_self` ever stashed its template, `_deduplicate_cache` would silently break it, and nothing in the current code would catch that.

---

## 1. The sharing rule is narrower than "same type, same cache size"

In practice a summand shares only if it is *simultaneously* (a) not behind any wrapper, (b) at the top level of the sum, and (c) structurally lucky. Three separate ways to fall out:

### (a) Any wrapper in between makes the cache invisible

`ScaledOperator` holds no cache of its own, so `getcache` returns `nothing` and the summand drops out of both paths — even though the cache it is hiding is one field deeper and perfectly shareable.

```julia
using SciMLOperators, LinearAlgebra, Random
Random.seed!(0)
n = 512
v = rand(n)
mk() = MatrixOperator(rand(n, n)) * MatrixOperator(rand(n, n))

bare()   = mk() + mk() + mk() + mk()
scaled() = 2.0*mk() + 3.0*mk() + 4.0*mk() + 5.0*mk()

cache_operator(bare(), v); cache_operator(scaled(), v)   # warm up

Lb, Ls = bare(), scaled()
@show @allocated cache_operator(Lb, v)
@show @allocated cache_operator(Ls, v)
```

```
bare   :  8608 bytes   # 2 buffers, shared across all 4 summands
scaled : 33520 bytes   # 8 buffers, one pair per summand
```

Same four terms, same cache layout, 3.9x the scratch. The only difference is a scalar coefficient.

It is not only `ScaledOperator`. Counting distinct buffers reachable from the whole tree (`n = 32`):

| sum | buffers | `getcache(ops[1])` |
|---|---|---|
| `A*B + C*D` | 2 | cache |
| `2*(A*B) + 3*(C*D)` | **4** | `nothing` |
| `A*B - C*D` | **4** | cache (but `ops[2]` is scaled) |
| `AffineOperator x2` | **4** | `nothing` |
| `InvertibleOperator x2` | **8** | `nothing` |
| `BlockDiagonalOperator x2` | **8** | `nothing` |

Note row 3: `A - B` lowers to `AddedOperator(A, -1*B)`, so **plain subtraction of two identically shaped terms gets zero sharing**. Between that and `α*A + β*B`, this covers most of what solver code actually builds.

(`adjoint`/`transpose` are fine, because `adjoint(::ComposedOperator)` returns another `ComposedOperator`. Nested sums are fine, because the `AddedOperator` constructor flattens them.)

### (b) Sharing stops at the top level of the sum

Caches owned *inside* a summand are never offered the corresponding cache in a sibling summand, even when the summands are the same concrete type with the same shape signature:

```
inv(A*B) + inv(C*D)      : 5 buffers
  InvertedOperator caches shared : true      # 1 buffer
  inner ComposedOperator shared  : false     # 2 + 2 buffers

(T⊗ * M) + (T⊗ * M)      : 4 buffers
  outer ComposedOperator shared  : true
  inner TensorProduct shared     : false
```

So the fraction of scratch actually shared falls off with tree depth, and the deep-tree case is exactly where the absolute numbers are largest.

### (c) `adopt_cache` for `TensorProductOperator` discards the donated cache

`cache_internals(::TensorProductOperator, v)` guards on `iscached(L)`, which is false whenever *any factor* is uncached — so it re-runs `cache_self` and throws away the cache it was just handed. And because `adopt_cache` returned non-`nothing`, `_cache_summand` returns early and `_deduplicate_cache` never gets a chance either:

```
TensorProductOperator(M(2,2)*M(2,2), M(4,4)*M(4,4)) summands :  shared = false
TensorProductOperator(M(2,2),        M(4,4))        summands :  shared = true
```

Direct confirmation of the mechanism:

```
after cache_self, iscached(T) = false
cache_self's buffer survived cache_internals: false    # composite factors
plain factors, iscached = true, buffer survives: true
```

Incidentally this means master already allocates that tensor cache twice on every `cache_operator` call for such an operator — pre-existing, but this PR is the reason to fix it.

### Suggested direction

The three are the same underlying issue: the opt-in keys off the concrete type of the *top-level* summand and requires the cache owner to be that exact object, so whether sharing happens depends on structural accidents rather than on whether the caches are actually interchangeable.

Narrow fixes, if you want to keep the current design:

```julia
# (a) forward through wrappers that hold no scratch of their own
getcache(L::ScaledOperator) = getcache(L.L)
update_cache(L::ScaledOperator, new_cache) = @reset L.L = update_cache(L.L, new_cache)
adopt_cache(L::ScaledOperator, cache, v) =
    (inner = adopt_cache(L.L, cache, v); inner === nothing ? nothing : (@reset L.L = inner))

# (c) match ComposedOperator's guard
function cache_internals(L::TensorProductOperator, v::AbstractVecOrMat)
    isnothing(L.cache) && (L = cache_self(L, v))
    ...
```

plus running `_deduplicate_cache` on the adopted result in `_cache_summand` as a belt-and-braces measure, so a mis-declared `adopt_cache` degrades to the post-hoc path rather than to nothing.

The type guard still does its job under (a): two `ScaledOperator{T,λType,LType}` summands have identical `LType`, so inner cache types and eltypes stay pinned exactly as they are today.

But it's worth deciding whether the wrapper-forwarding rule should be general rather than added type by type — otherwise the next transparent wrapper reintroduces the hole silently, with no test that would notice. The fully general version of this is a recursive walk over each summand's tree with a pool of retired buffers, rather than an opt-in on the top-level type; that's a bigger change and probably a follow-up, but it's the thing the current design is approximating.

## 2. Compile time is superlinear in the number of summands

`_deduplicate_cache`/`_try_adopt` specialize over every suffix of the `earlier`/`donors` tuples, so it's O(N^2) instantiations. First `cache_operator` call on a sum of same-typed composed operators (`n = 64`):

| summands | master | PR |
|---|---|---|
| 4 | 0.041s | 0.165s |
| 8 | 0.075s | 0.313s |
| 16 | 0.130s | 0.726s |
| 32 | 0.285s | **2.564s** |

Package precompile also goes 2.5s -> 3.9s. It's one-time and bounded, but large-term Hamiltonians are the target use case here, so it deserves a mention at minimum.

## Smaller points

- **Not declared `public`.** `getcache`, `update_cache` and `adopt_cache` are documented extension hooks, but they aren't in the `Expr(:public, ...)` block at `src/SciMLOperators.jl:329-340` — unlike `has_tensor_outer_mul_fast`/`tensor_outer_mul_fast!`, which are the existing precedent for exactly this kind of hook. Downstream code overloading them won't pass ExplicitImports' public-API check.
- **No version bump.** `Project.toml` stays at 1.25.2. This adds public API, so it should be 1.26.0 — both #413 and #416 bumped in-PR.
- **`_slots_match` ignores aliasing.** It compares slot sizes but not which slots are `===` each other. Today that's safe by accident rather than by construction: `TensorProductOperator` is the only type with intentionally aliased slots, and `c5 === c1` iff both factors are square, which is exactly the condition under which `size(c1) == size(c5)`. If a future cache breaks that coincidence, a non-aliased operator could adopt an aliased cache and silently corrupt itself. Comparing the aliasing pattern explicitly is cheap and makes the invariant hold by construction.
- **Benchmark framing.** The win scales with the cache-to-operator ratio rather than universally. I reproduced the sparse setup at 10 summands with a *vector* input: 2.68 MB -> 2.09 MB (~22%). With dense `n = 64` operators it's 1.6%. The plot in the description uses `u = rand(N, N)`, which makes the caches `N x N` and therefore dominant — legitimate, but the axis label doesn't say the input is a matrix, and the log-log scale reads as a much larger general win than the vector case delivers.
- **Thread safety** deserves one line in the docs: a cached operator was never safe to apply concurrently, but now the summands of an `AddedOperator` are not independently safe either. Anyone hand-parallelizing over `L.ops` would be newly broken.

Not covered on my side: GPU paths and the docs build (`linkcheck = true`).

albertomercurio:
Thanks here what I fixed:

- Transparent wrappers (`ScaledOperator`, `AdjointOperator`, `TransposedOperator`) forward the hooks now, so `2A+3B+4C+5D` goes 33 344 B → 8 432 B, same as the unscaled sum. `A - B` shares its scratch too, which it previously didn't.
- `cache_internals(::TensorProductOperator, v)` guards on `isnothing(L.cache)` instead of the recursive `iscached(L)`, so a donated cache isn't discarded. This is a pre-existing bug — on master such an operator allocates its cache twice. Happy to split it out if you'd rather it landed separately.
- An `adopt_cache` that returns a cache other than the one it was handed is treated as having declined, so it degrades to the post-hoc path. Cheaper than deduplicating the adopted result, which cost ~0.5 s at 32 summands.
- `_slots_match` compares the aliasing pattern; hooks declared `public`. Left the version alone for you to set.

PR #382, opened 2026-05-17
Remove SafeTestsets dependency

visr:
It is still a test dependency.
Introduced recently in #372.


oscardssmith:
@ChrisRackauckas can you release?

JamesWrigley:
Bump, would be great to get this in a release.

ChrisRackauckas:
Huh the registration stalled for some reason, retrying it.

PR #387, opened 2026-06-07
feat: Implement lazy evaluation flag for WOperator to optimize memory and computation

swiatlowemnie333:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
- [ ] Any new documentation only uses public API

### DIFF
--- a/src/woperator.jl
+++ b/src/woperator.jl
@@ -54,8 +54,9 @@ mutable struct WOperator{
     _func_cache::F           # cache used in `mul!`
     _concrete_form::C        # non-lazy form (matrix/number) of the operator
     jacvec::JV
+    lazy::Bool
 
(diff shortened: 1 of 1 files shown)


swiatlowemnie333:
​Yo so about that commit i did for WOperator in SciMLOperators i added a lazy flag cuz i needed it for optimising the feedback loop in the system. i changed the mutable struct WOperator so it stores a lazy::Bool and updated the constructor so its false by default so it dont break old code. also fixed the copy method so it transfers the state right. all 947 tests passed and are green so its working good and no errors. i think its realy needed to manage memory and not dump everything at once when not needed. Peace.
Zazul

ChrisRackauckas:
It doesn't seem to be used anywhere?

PR #397, opened 2026-06-30
WOperator{!IIP}: materialize MatrixOperator mass matrix in convert(AbstractMatrix, W)

### LINKED ISSUE #396: WOperator{!IIP}: convert(AbstractMatrix, W) fails when mass_matrix isa MatrixOperator
### Summary

For an out-of-place `WOperator` whose `mass_matrix` is a `MatrixOperator` (or, more generally, any `AbstractSciMLOperator`), calling `convert(AbstractMatrix, W)` errors with a `MethodError` instead of returning the materialized `-M/γ + J`. This breaks every downstream solver path that materializes `W` lazily — e.g. `OrdinaryDiffEqNonlinearSolve`'s OOP `NLNewton`, which calls `W \ b` (which in turn calls `convert(AbstractMatrix, W) \ b`).

### Minimal repro

```julia
using SciMLOperators
using SciMLOperators: MatrixOperator, WOperator

function update_mm!(A, u, p, t)
    A[1, 1] = cos(t); A[2, 2] = cos(t)^2 + u[3]; A[3, 3] = t*cos(t) + 1
    A[1, 2] = sin(t); A[2, 1] = sin(t)*u[1]; A
end
update_mm(A, u, p, t) = (B = copy(A); update_mm!(B, u, p, t); B)

n = 3
mm = MatrixOperator(ones(n, n); update_func = update_mm, update_func! = update_mm!)
J = randn(n, n)
W = WOperator{false}(mm, 0.1, J, randn(n))

convert(AbstractMatrix, W)  # boom
```

```
ERROR: MethodError: Cannot `convert` an object of type
SciMLOperators.AddedOperator{Float64, Tuple{
  SciMLOperators.ScaledOperator{Float64, ScalarOperator{...}, MatrixOperator{Float64, Matrix{Float64}, ...}},
  MatrixOperator{Float64, Matrix{Float64}, ...}}}
to an object of type Matrix{Float64}
```

### Root cause

[`src/woperator.jl:124-130`](https://github.com/SciML/SciMLOperators.jl/blob/master/src/woperator.jl#L124-L130):

```julia
function Base.convert(::Type{AbstractMatrix}, W::WOperator{IIP}) where {IIP}
    if !IIP
        # Allocating
        W._concrete_form = -W.mass_matrix / W.gamma + convert(AbstractMatrix, W.J)
    end
    return W._concrete_form
end
```

`W.J` is materialized via `convert(AbstractMatrix, ...)`, but `W.mass_matrix` is left in operator form. So `-MatrixOperator / Number` becomes `ScaledOperator`, then `ScaledOperator + Matrix` becomes `AddedOperator`, which then can't be assigned back to the `_concrete_form` slot — that slot is typed `Matrix{Float64}` (set by the constructor at [`src/woperator.jl:62-67`](https://github.com/SciML/SciMLOperators.jl/blob/master/src/woperator.jl#L62-L67), which *does* materialize the mass matrix):

```julia
AJ = J isa MatrixOperator ? convert(AbstractMatrix, J) : J
mm = mass_matrix isa MatrixOperator ?
    convert(AbstractMatrix, mass_matrix) : mass_matrix
_concrete_form = -mm / gamma + AJ
```

So the constructor and the refresh path are inconsistent: the constructor materializes the mass matrix; the refresh path does not.

### Impact

This is the underlying cause of [SciML/OrdinaryDiffEq.jl#XXXX] — the OOP `NLNewton` "Dependent Mass Matrix Tests" failures on master. Any user combining `iip = false` + `mass_matrix = MatrixOperator(...)` + an implicit solver that ends up calling `convert(AbstractMatrix, W)` (essentially: anything routed through OOP `WOperator \ b`) hits this.

### Proposed fix

Mirror the constructor's materialization in the refresh path:

```julia
function Base.convert(::Type{AbstractMatrix}, W::WOperator{IIP}) where {IIP}
    if !IIP
        mm = W.mass_matrix isa MatrixOperator ?
            convert(AbstractMatrix, W.mass_matrix) : W.mass_matrix
        W._concrete_form = -mm / W.gamma + convert(AbstractMatrix, W.J)
    end
    return W._concrete_form
end
```

Verified locally: repro above passes; `W \ b` produces a residual of `~1e-15`; `update_coefficients!(W, u, p, t; gamma)` followed by `convert(AbstractMatrix, W)` returns a fresh, correctly-updated matrix.

I'll open a PR with this fix plus a regression test.

### Notes

- The same `isa MatrixOperator` check is what the constructor uses; broader `AbstractSciMLOperator` handling would be a follow-up.
- The IIP path is unaffected — `convert(::Type{AbstractMatrix}, W::WOperator{true})` is a no-op since `_concrete_form` is mutated externally via `jacobian2W!`.
- Bug present on at least v1.15.1 through v1.22.1 (every version I have cached locally).

singhharsh1708:
Fixes #396.

`Base.convert(::Type{AbstractMatrix}, W::WOperator{!IIP})` re-materializes `_concrete_form` from `W.mass_matrix` and `W.J`, but only `W.J` was being passed through `convert(AbstractMatrix, ...)`. When `W.mass_matrix` is a `MatrixOperator`, `-W.mass_matrix / W.gamma` stays an `AbstractSciMLOperator` (`ScaledOperator`), then `+ convert(AbstractMatrix, W.J)` becomes an `AddedOperator`, which cannot be assigned back to the `Matrix`-typed `_concrete_form` slot. Result: a `MethodError`.

The constructor at `src/woperator.jl:62-67` already materializes the mass matrix:
```julia
mm = mass_matrix isa MatrixOperator ?
    convert(AbstractMatrix, mass_matrix) : mass_matrix
_concrete_form = -mm / gamma + AJ
```
This PR mirrors that same conversion in the refresh path. No other paths change.

Added a regression test under `test/shared/WOperator.jl` that:
- Builds an OOP `WOperator` with a `MatrixOperator` mass matrix
- Calls `convert(AbstractMatrix, W)` and verifies the value matches the reference
- Re-materializes after `update_coefficients!(W, u, p, t; gamma)`
- Verifies `W \ b` works end-to-end

## Why it matters

This is the proximate cause of OrdinaryDiffEq.jl's OOP `NLNewton` "Dependent Mass Matrix Tests" failing on master — any user combining `iip=false` + `mass_matrix = MatrixOperator(...)` + an implicit solver hits this. With this fix in, the OrdinaryDiffEq side can drop the `error("Non-concrete Jacobian not yet supported by out-of-place Newton solve.")` for the SciMLOperator-mass-matrix path. Follow-up PR there will bump the `SciMLOperators` compat floor once this is tagged.

## Test plan
- [x] New `test/shared/WOperator.jl` regression covers the MatrixOperator + OOP + `\` path
- [x] Existing WOperator tests still pass

### DIFF
--- a/src/woperator.jl
+++ b/src/woperator.jl
@@ -123,8 +123,12 @@ end
 
 function Base.convert(::Type{AbstractMatrix}, W::WOperator{IIP}) where {IIP}
     if !IIP
-        # Allocating
-        W._concrete_form = -W.mass_matrix / W.gamma + convert(AbstractMatrix, W.J)
+        # Mirror the constructor: materialize a MatrixOperator mass matrix so the
+        # result type matches `_concrete_form` (otherwise this becomes an
+        # AddedOperator and the assignment back into the Matrix-typed slot fails).
+        mm = W.mass_matrix isa MatrixOperator ?
+            convert(AbstractMatrix, W.mass_matrix) : W.mass_matrix
+        W._concrete_form = -mm / W.gamma + convert(AbstractMatrix, W.J)
     end
     return W._concrete_form
 end
--- a/test/shared/WOperator.jl
+++ b/test/shared/WOperator.jl
@@ -27,4 +27,28 @@ Random.seed!(0)
     W_scalar2 = WOperator{false}(I, 0.5, ScalarOperator(2.0), 1.0)
     @test_deprecated update_coefficients!(W_scalar2; dtgamma = 0.25)
     @test convert(Number, W_scalar2) ≈ 2.0 - 1.0 / 0.25
+
+    # OOP WOperator with a MatrixOperator mass matrix: convert(AbstractMatrix, W)
+    # must materialize the mass matrix so the result fits in the
+    # Matrix-typed _concrete_form slot (issue #396).
+    n = 3
+    mm_data = randn(n, n)
+    update_mm!(A, u, p, t) = (A[1, 1] = cos(t); A[2, 2] = u[1]; A)
+    update_mm(A, u, p, t) = (B = copy(A); update_mm!(B, u, p, t); B)
+    mm_op = MatrixOperator(
+        copy(mm_data); update_func = update_mm, update_func! = update_mm!,
+    )
+    J_mm = randn(n, n)
+    u_mm = randn(n)
+    gamma_mm = 0.1
+    W_mm = WOperator{false}(mm_op, gamma_mm, J_mm, u_mm)
+    W_ref = -convert(AbstractMatrix, mm_op) / gamma_mm + J_mm
+    @test convert(AbstractMatrix, W_mm) ≈ W_ref
+    # Re-materialization after update_coefficients!
+    update_coefficients!(W_mm, randn(n), nothing, 1.5; gamma = 0.2)
+    W_ref2 = -convert(AbstractMatrix, W_mm.mass_matrix) / 0.2 + J_mm
+    @test convert(AbstractMatrix, W_mm) ≈ W_ref2
+    # And `\` works end-to-end
+    b = randn(n)
+    @test W_mm \ b ≈ convert(AbstractMatrix, W_mm) \ b
 end


singhharsh1708:
@ChrisRackauckas

PR #413, opened 2026-07-26
Allocate only the TensorProductOperator cache slots a vector input can reach

albertomercurio:
## Description

`cache_self(::TensorProductOperator, v)` allocates all seven cache slots regardless of the shape of `v`. Five of them — `c2`, `c3`, `c4`, `c6`, `c7` — are read only from the `k > 1` branches of `outer_mul!` and `outer_div!`, and **every path that reaches them returns early when the input has a single column**:

| slot | only read at | guarded by |
|---|---|---|
| `c2`, `c3` | `outer_mul!` 3-arg, `tensor.jl:704` | `if k == 1 … return w` at `:692` |
| `c2`, `c3`, `c4` | `outer_mul!` 5-arg, `tensor.jl:747` | `if k == 1 … return w` at `:735` |
| `c6`, `c7` | `outer_div!`, `tensor.jl:813` | `if k == 1 … return v` at `:806` |
| `c1` | 3/5-arg `mul!`, `:523` and `:558` | always reachable |
| `c5` | 3-arg `ldiv!`, `:592` | always reachable |

(Line numbers are as of this branch.)

So a vector input only ever touches `c1` and `c5`, and `c5 === c1` already whenever both factors are square. The remaining buffers are allocated, zeroed by `lmul!(false, …)`, and never read again.

This PR splits `cache_self` on `AbstractVector`/`AbstractMatrix` and allocates only the reachable slots in the vector case. Nothing about the matrix path changes, and no buffer is shared between operators — the change only stops allocating buffers that provably cannot be reached.

Dispatching on the input type rather than branching on `size(v, 2)` is deliberate: a runtime branch would make the cache tuple a `Union` of two shapes and cost inference on `cache_operator`'s return type.

## Numbers

Distinct buffers held in the cache and their total `sizeof`, measured on `master` vs this branch (script at the bottom).

**Square factors, vector input**

| operator | `length(v)` | before | after | |
|---|---|---|---|---|
| 4×4 ⊗ 4×4 | 16 | 4 buffers, 512 B | 1 buffer, 128 B | **−75%** |
| 16×16 ⊗ 16×16 | 256 | 4 buffers, 8 192 B | 1 buffer, 2 048 B | **−75%** |
| 64×64 ⊗ 64×64 | 4 096 | 4 buffers, 131 072 B | 1 buffer, 32 768 B | **−75%** |
| 256×256 ⊗ 256×256 | 65 536 | 4 buffers, 2 097 152 B | 1 buffer, 524 288 B | **−75%** |

**Non-square factors, vector input**

| operator | `length(v)` | before | after | |
|---|---|---|---|---|
| 5×3 ⊗ 7×11 | 33 | 7 buffers, 2 040 B | 2 buffers, 608 B | **−70%** |
| 40×24 ⊗ 56×88 | 2 112 | 7 buffers, 130 560 B | 2 buffers, 38 912 B | **−70%** |
| 100×60 ⊗ 140×220 | 13 200 | 7 buffers, 816 000 B | 2 buffers, 243 200 B | **−70%** |

**Matrix input — unchanged, as intended**

| operator | input | before | after |
|---|---|---|---|
| 64×64 ⊗ 64×64 | `(4096, 10)` | 4 buffers, 1 310 720 B | 4 buffers, 1 310 720 B |

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/src/interface.jl
+++ b/src/interface.jl
@@ -184,7 +184,7 @@ cache_internals(L::AbstractSciMLOperator, ::AbstractVecOrMat) = L
 ###
 
 Base.size(A::AbstractSciMLOperator, d::Integer) = d <= 2 ? size(A)[d] : 1
-Base.eltype(::Type{AbstractSciMLOperator{T}}) where {T} = T
+Base.eltype(::Type{<:AbstractSciMLOperator{T}}) where {T} = T
 Base.eltype(::AbstractSciMLOperator{T}) where {T} = T
 Base.promote_eltype(::AbstractSciMLOperator{<:T1}, ::AbstractSciMLOperator{<:T2}) where {T1, T2} = Base.promote_type(T1, T2)
 
--- a/src/tensor.jl
+++ b/src/tensor.jl
@@ -426,7 +426,37 @@ function Base.:\(L::TensorProductOperator, v::AbstractVecOrMat)
     return v isa AbstractMatrix ? reshape(V, (n, k)) : reshape(V, (n,))
 end
 
-function cache_self(L::TensorProductOperator, v::AbstractVecOrMat)
+# `c1` (3 and 5 arg `mul!`) and `c5` (3 arg `ldiv!`) are read for every input shape.
+function _cache_self_c1_c5(L::TensorProductOperator, v::AbstractVecOrMat)
+    outer, inner = L.ops
+
+    mi, ni = size(inner)
+    mo, no = size(outer)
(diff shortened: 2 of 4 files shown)


albertomercurio:
Hello, I think this PR is relatively easy to review, it only reshapes the code without changing the functionalities.

ChrisRackauckas (approved):

PR #419, opened 2026-08-15
Give WOperator a Jacobian staleness flag

ChrisRackauckas-Claude:
> **Draft — please ignore until reviewed by @ChrisRackauckas.**

## What and why

A solver that caches a factorization of `W.J` across steps has to tell two events apart:

- **`gamma` moved** — an `O(n²)` update to the cached factorization, if the algorithm supports one.
- **`J`'s contents changed** — an `O(n³)` refactorization, unavoidably.

`gamma` it can compare for itself, because it is a field. An in-place write to `J` — which is exactly what `calc_J!` does — is invisible from the operator: same object, same size, same everything. There is currently no way to ask.

This adds a `jac_version::Int` to `WOperator`, bumped by `mark_jacobian_updated!` and read by `jacobian_version`. Both no-op on anything that isn't a `WOperator` (returning the argument and `0` respectively), so a caller can use them without a type check.

The motivating consumer is SciML/LinearSolve.jl#1215, which reduces `J` once to Hessenberg form and absorbs each new `gamma` in `O(n²)`; without this it cannot distinguish the two cases and has to reduce every step, which is strictly worse than an LU. But the interface is not specific to that algorithm — any caching factorization of a `WOperator` needs the same bit.

## Verification

`Pkg.test()` on the branch:

```
Testing SciMLOperators tests passed
```

New testset in `test/shared/WOperator.jl` (11 assertions): the counter starts at 0, `update_coefficients!(W; gamma)` does *not* bump it, an explicit `mark_jacobian_updated!` does, `copy` carries it, and the no-op methods work off the type.

Writing that test surfaced something worth stating explicitly, since it is easy to assume otherwise: for the in-place, plain-matrix case `Matrix(W)` returns the externally-maintained `_concrete_form`, so after an in-place write to `J` it is **deliberately stale** until the owner (OrdinaryDiffEq's `jacobian2W!`) refreshes it. The test asserts that staleness rather than papering over it — it is the same invisibility this counter exists to cover.

## Cost

One `Int` per `WOperator`. No behavior change: nothing reads `jac_version` unless it opts in.

## Judgment calls

- **A counter, not a dirty flag.** A flag needs someone to clear it, and there is no single consumer to decide when; a monotonic counter lets any number of independent caches each compare against the generation *they* last saw.
- **`mark_jacobian_updated!(A) = A` for the generic fallback** so integrator code can call it unconditionally rather than branching on the `W` type.
- **Minor version bump** (1.26.1 → 1.27.0) for the new public API.

## Not verified

Downstream packages beyond OrdinaryDiffEq's `calc_W!` path, GPU paths, the docs build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014CrTAi2YLkosEXXbudXyKf

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "SciMLOperators"
 uuid = "c0aeaf25-5076-4817-a8d5-81caf7dfa961"
-version = "1.26.1"
+version = "1.27.0"
 authors = ["Vedant Puri <vedantpuri@gmail.com>"]
 
 [deps]
--- a/src/SciMLOperators.jl
+++ b/src/SciMLOperators.jl
@@ -303,7 +303,9 @@ export
     TensorProductOperator,
     TensorSumOperator,
     WOperator,
-    StaticWOperator
+    StaticWOperator,
+    jacobian_version,
+    mark_jacobian_updated!
 
 export update_coefficients!,
     update_coefficients, isconstant,
--- a/src/SciMLOperators.jl
+++ b/src/SciMLOperators.jl
@@ -304,8 +304,9 @@ export
     TensorSumOperator,
     WOperator,
     StaticWOperator,
-    jacobian_version,
-    mark_jacobian_updated!
+    jacobian_stale,
+    mark_jacobian_updated!,
+    mark_jacobian_current!
 
 export update_coefficients!,
     update_coefficients, isconstant,
--- a/src/woperator.jl
+++ b/src/woperator.jl
@@ -122,6 +122,11 @@ mutable struct WOperator{
     _func_cache::F           # cache used in `mul!`
     _concrete_form::C        # non-lazy form (matrix/number) of the operator
     jacvec::JV
+    # Bumped by `mark_jacobian_updated!` whenever `J`'s contents change. A solver that
+    # caches a factorization of `J` across steps needs to tell a new `gamma` from a new
+    # Jacobian; `gamma` it can compare directly, but an in-place write to `J` is otherwise
+    # invisible. See `jacobian_version`.
+    jac_version::Int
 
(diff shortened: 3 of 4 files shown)


ChrisRackauckas-Claude:
**Switched from a generation counter to a `jac_stale::Bool`** per review preference. API is now:

- `jacobian_stale(W)` — predicate, conservatively `true` off the type so a caller that cannot ask just refactorizes
- `mark_jacobian_updated!(W)` — set it (unchanged)
- `mark_jacobian_current!(W)` — clear it, called by the consumer after it refactorizes

Simpler, and one `Bool` instead of an `Int`. The tradeoff is real and documented on `mark_jacobian_current!`: the flag is **shared, not per-consumer**, so two solvers caching factorizations of the same `WOperator` will race — whichever clears first hides the event from the other. One consumer per operator.

`Pkg.test()` green; the testset now also pins that `mark_jacobian_updated!` is idempotent (which a counter was not) and that the untracked fallback is conservative.

---

Worth recording why this interface is needed at all, since "just track whether `gamma` changed" is the natural first thought. Measured on a Brusselator at n=800, tol 1e-8, 800 steps:

| transition | count |
|---|---|
| `gamma` and `J` both changed | **2** |
| `gamma` changed, `J` did not | 96 |
| `J` changed, `gamma` did not | **0** |
| neither changed | 702 |

A `gamma`-changed signal fires on 98 steps and cannot separate the 2 that matter. Every Jacobian recompute was masked by a simultaneous `gamma` change — `calc_W!` recomputes `J` because Newton is struggling, and the controller has just moved `dt` for the same reason, so they co-occur by construction. And `gamma` needs no tracking regardless: it is a field, so a consumer just compares it. The flag is for the event with no observable signature — an in-place `J .= …` leaves the same object, same size, same everything.

PR #427, opened 2026-08-26
Follow a WOperator's Jacobian in has_concretization

AJ0070:
`has_concretization(::AbstractWOperator)` was an unconditional `true`, so a `WOperator` over a matrix-free Jacobian claimed a concrete form it cannot produce and callers that guard on the trait still threw from `convert`. Every other wrapping operator here already follows what it wraps, so this does the same and ANDs the mass matrix with the Jacobian.

Left the abstract method returning `true` and put the new one on `WOperator`, since a subtype like the `GenericWOperator` in the tests has no `mass_matrix` or `J` to ask.

Found from SciML/LinearSolve.jl#1236, where the default solver builds a cacheval for every slot and the concretizing ones threw a `MethodError` out of `init`.

AI Disclosure: Used Opus 5


src/woperator.jl
```diff
@@ -481,3 +481,13 @@ function LinearAlgebra.mul!(Y::AbstractArray, W::WOperator, B::AbstractArray)
 end
 
 has_concretization(::AbstractWOperator) = true
+
+# A `WOperator` is only as concretizable as the pieces it holds, the same way a
+# `ScaledOperator` is. Answering an unconditional `true` made one over a matrix-free
+# Jacobian claim a concrete form it cannot produce, so callers that guard on this trait
+# still threw from `convert`. Left on the abstract type for subtypes that carry their
+# concrete form some other way.
+# See https://github.com/SciML/LinearSolve.jl/issues/1236.
```
ChrisRackauckas:
```suggestion
```

